PR #5, opened 2024-10-21
Implent API for setting user info and assigning user ID

limbonaut:
This PR exposes API to GDScript to work with sentry user information. 

Key points:
- Set/get user info.
- If not set explicitly, a user ID (UUID) is generated automatically.
- Option to infer IP address (function call).
- User information is stored in a runtime config (a file in the user data directory).
- Single-header UUID library is added as a Git submodule.

### DIFF
--- a/project/main.gd
+++ b/project/main.gd
@@ -19,6 +19,7 @@ var _event_level: Sentry.Level
 func _ready() -> void:
 	level_choice.get_popup().id_pressed.connect(_on_level_choice_id_pressed)
 	_init_level_choice_popup()
+	_init_user_info()
 
 
 func _init_level_choice_popup() -> void:
@@ -30,6 +31,15 @@ func _init_level_choice_popup() -> void:
 	popup.add_item("FATAL", Sentry.LEVEL_FATAL + 1)
 
 	_on_level_choice_id_pressed(Sentry.LEVEL_INFO + 1)
+
+
+func _init_user_info() -> void:
+	# The user info is persisted in the user data directory (referenced by "user://"),
+	# so it will be loaded again on subsequent launches.
+	var user: SentryUser = Sentry.get_user()
+	username.text = user.username
+	email.text = user.email
+	user_id.text = user.user_id
 
 
 func _on_level_choice_id_pressed(id: int) -> void:
--- a/project/main.gd
+++ b/project/main.gd
@@ -93,11 +93,15 @@ func _on_set_context_pressed() -> void:
 
 
 func _on_set_user_button_pressed() -> void:
-	print("Setting user info.")
+	print("Setting user info...")
 	var sentry_user := SentryUser.new()
-	sentry_user.user_id = user_id.text
+	if not user_id.text.is_empty():
+		sentry_user.user_id = user_id.text
+	else:
+		sentry_user.assign_unique_id()
 	sentry_user.username = username.text
 	sentry_user.email = email.text
 	if infer_ip.button_pressed:
 		sentry_user.infer_ip_address()
 	Sentry.set_user(sentry_user)
+	print("   ", sentry_user)
--- a/project/main.gd
+++ b/project/main.gd
@@ -8,6 +8,10 @@ extends CanvasLayer
 @onready var tag_value: LineEdit = %TagValue
 @onready var context_name: LineEdit = %ContextName
 @onready var context_expression: CodeEdit = %ContextExpression
+@onready var user_id: LineEdit = %UserID
+@onready var username: LineEdit = %Username
+@onready var email: LineEdit = %Email
+@onready var infer_ip: CheckBox = %InferIP
 
 var _event_level: Sentry.Level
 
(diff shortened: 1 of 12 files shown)


limbonaut:
PR is ready for a review.

project/main.gd
```diff
@@ -28,6 +33,15 @@ func _init_level_choice_popup() -> void:
 	_on_level_choice_id_pressed(Sentry.LEVEL_INFO + 1)
 
 
+func _update_user_info() -> void:
+	# The user info is persisted in the user data directory (referenced by "user://"),
+	# so it will be loaded again on subsequent launches.
+	var user: SentryUser = Sentry.get_user()
+	username.text = user.username
+	email.text = user.email
+	user_id.text = user.user_id
```
bruno-garcia: nit: Not related to the peristance but the user property is usually called `user.id` (since user.user_id feels redundant):
```suggestion
	user_id.text = user.id
```

src/sentry_user.h
```diff
@@ -0,0 +1,43 @@
...
+	void set_username(const String &p_username) { username = p_username; }
+	String get_username() const { return username; }
+
+	void set_email(const String &p_email) { email = p_email; }
+	String get_email() const { return email; }
+
+	void set_ip_address(const String &p_ip_address) { ip_address = p_ip_address; }
+	String get_ip_address() const { return ip_address; }
+
+	void infer_ip_address() { ip_address = "{{auto}}"; }
```
bruno-garcia, line 36: This function would benefit from code docs for the caller.

src/sentry_user.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+#include <godot_cpp/core/error_macros.hpp>
+#include <godot_cpp/variant/packed_string_array.hpp>
+#include <godot_cpp/variant/variant.hpp>
+
+bool SentryUser::is_user_valid() const {
+	return !user_id.is_empty() || !username.is_empty() || !email.is_empty() || !ip_address.is_empty();
+}
+
+String SentryUser::_to_string() const {
```
bruno-garcia, line 13: Is this used? doesn't seem to be called from anywhere?

src/sentry_singleton.cpp
```diff
@@ -457,6 +502,9 @@ Sentry::Sentry() {
 		return;
 	}
 
+	// Load the runtime configuration from the user's data directory.
+	runtime_config.load_file(OS::get_singleton()->get_user_data_dir() + "/sentry.dat");
```
bruno-garcia, line 506: for testing things while mocking file I/O, there's a way to hook into what OS::get_singleton returns? I imagine we wouldn't want concurrent tests runnign racing over files

src/sentry_user.h
limbonaut: Godot has a built-in documentation system. I think, explanations would need to go there, because that's where most users would search for answers. 
image
I'll tackle built-in docs in a separate PR, since it needs some code to be added in extension.

project/main.gd
limbonaut: Corrected

src/sentry_user.cpp
limbonaut: Yes, it is used to print user info in the demo project:
```gdscript
	Sentry.set_user(sentry_user)
	print("   ", sentry_user)
```
Typically, you expect something useful from `print(sentry_user)` in GDScript - this function provides it.

src/sentry_singleton.cpp
limbonaut: There is an option in the project settings to use a custom user dir, but I don't think it would be useful here. Could implement a CLI option.

src/sentry_user.h
bruno-garcia: Got it. Would be cool if that tool pulled docs from code. So code and docs could live at the same location

src/sentry_user.cpp
bruno-garcia: If it's just used on the demo project, I'd suggest putting the function there to simplify the API surface of the SDK. But if this is an expectation of devs to debug in Godot, makes sense to keep it here.

src/sentry_singleton.cpp
bruno-garcia: Not sure either of those would help with concurrent tests and unit tests. 
In the unit test case, you'd want to mock that and setup different behavior (what if it throws, or returns A or B etc) to verify how the code calling it behaves.

In case of integration tests you'd at least want to make sure that each test has a single/random file folder/path to work on. So that they don't interact with each other. And at the end of the test run the root directory where all the tests access under, would be deleted, for clean up.

bruno-garcia:
> Advantage is that it would execute code via GDScript - just like users would

This is nice. Can serve as API usage examples perhaps too. Makes it easier to see if we're breaking the API too if we don't have snapshot testing for the public API (we avoid breaking public API in general. If we must, we do in major releases)

> Another possibility is one of the C++ testing frameworks. Speaking of which, Godot Engine already uses C++ doctest framework - such tests could be added for any engine module. If sentry SDK is to be compiled as module, we could add tests using that framework for the module code, like this https://github.com/limbonaut/limboai/tree/master/tests
Advantage - no extra dependencies needed. But this SDK can't be compiled as an engine module right now - it would require some work.

Sounds like we need a combination of both? Or what would you recommend?

bruno-garcia (approved):
Approving to unblock, early days so code doc and tests can follow later once things are more set up here

src/sentry_user.h
limbonaut: That tool is part of Godot, e.g. `godot --doctool`. It generates XML files for API exposed to `ClassDB`, and also updates those files on subsequent runs, but it only deals with the structure. The descriptions need to be added manually to those XML files. It preserves the added content on subsequent runs. There is also a tool to export sphinx-based RST-formatted class documentation.

Issue #4, opened 2024-10-16
Report Godot errors with a script stack trace


limbonaut:
It would be nice to have errors reported with a timestamp that happened prior to a crash or event.

![Image](https://github.com/user-attachments/assets/e7ba4489-7aa0-4ff4-9e96-d84ab5912a1a)

Logging is managed by https://github.com/godotengine/godot/blob/master/core/io/logger.h
This functionality is not exposed in Godot API, and as such cannot be used in GDExtension/GDScript.

AFAIK, there are two PRs possibly landing in Godot that expose logging:
  - https://github.com/godotengine/godot/pull/87576   
  - https://github.com/godotengine/godot/pull/91006

For the time being, we can explore a log-based parser and see how performant such approach can be.

UPDATE:
- PR #201

limbonaut:
We have intermediate solution in place. Waiting for logging to be exposed in the Godot Engine API.
- https://github.com/godotengine/godot/pull/91006

limbonaut:
Required changes were merged in Godot:
- https://github.com/godotengine/godot/pull/91006

Finally, proper logger support. I'm working on integrating these changed in [godot-4.5-support](https://github.com/getsentry/sentry-godot/tree/feat/godot-4.5-support) branch. This will bring script stack traces and better GDScript error events.

<img width="839" alt="Image" src="https://github.com/user-attachments/assets/c9e8ab04-8341-45ea-b931-ca54bf2cd5ad" />


We'll probably have to bump the minimum Godot version supported to 4.5. These changes should land with 1.0 version:
- #173

limbonaut:
This functionality will be released with `1.0-alpha.0`.

limbonaut:
Landed with preview `1.0.0-alpha.0`.

PR #9, opened 2024-10-30
Intermediary Error logger implementation

limbonaut:
While we are waiting for the Godot PRs (see #4), this is a logger implementation based on log file parsing. For practical usage, my performance measurements showed it as quite usable, and it's completely optional for game developers. It should be trivial to adapt it when the user-facing logger class is exposed in the Godot public API.

The current limitation is that we don't have access to a script stack trace, but we still have the line number, the function and the file in which the error has occurred (almost equivalent to the top frame of the stack trace).

### Error logger section in Project Settings
image

### Error captured as event
image
Note: the source code is automatically included with the event if the corresponding option is enabled. It is not available if project scripts are exported as binary tokens.

### DIFF
--- a/project/project.godot
+++ b/project/project.godot
@@ -30,4 +30,4 @@ window/stretch/aspect="keep_height"
 
 config/dsn="https://3f1e095cf2e14598a0bd5b4ff324f712@o447951.ingest.us.sentry.io/6680910"
 config/debug=true
-config/error_logger_capture_type=1
+config/error_logger/capture_type=1
--- a/project/project.godot
+++ b/project/project.godot
@@ -26,3 +26,4 @@ window/stretch/aspect="keep_height"
 
 config/dsn="https://3f1e095cf2e14598a0bd5b4ff324f712@o447951.ingest.us.sentry.io/6680910"
 config/debug=true
+config/error_logger_capture_type=1
--- a/project/project.godot
+++ b/project/project.godot
@@ -15,6 +15,10 @@ config/version="0.1.0"
 run/main_scene="res://main.tscn"
 config/features=PackedStringArray("4.3")
 
+[debug]
+
+file_logging/enable_file_logging=true
+
 [display]
 
 window/size/viewport_width=1440
--- a/src/sentry_options.h
+++ b/src/sentry_options.h
@@ -8,6 +8,12 @@
 #include <godot_cpp/variant/variant.hpp>
 
 class SentryOptions {
+public:
+	enum CaptureType {
+		CAPTURE_AS_BREADCRUMB,
+		CAPTURE_AS_EVENT,
+	};
+
 private:
 	static SentryOptions *singleton;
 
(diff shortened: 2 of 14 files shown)


limbonaut:
I've spun off a branch from this PR's - going to wait for the review/merge, and later rebase onto main.
I know it's risky, but it's either wait or have even bigger rebase later on.

project/project.godot
```diff
@@ -15,6 +15,10 @@ config/version="0.1.0"
 run/main_scene="res://main.tscn"
 config/features=PackedStringArray("4.3")
 
+[debug]
+
+file_logging/enable_file_logging=true
```
bruno-garcia, line 20: How about this as a name?

```suggestion
file_logging/enable_log_file_reader=true
```

src/sentry_util.h
```diff
@@ -1,16 +1,45 @@
...
+	static sentry_value_t variant_to_sentry_value(const Variant &p_variant);
+	static CharString get_screen_orientation_cstring(int32_t p_screen = -1);
 	static void sentry_event_set_context(sentry_value_t p_event, const char *p_context_name, sentry_value_t p_context);
 	static godot::CharString generate_uuid();
+	static sentry_value_t strings_to_sentry_list(const PackedStringArray &p_strings);
+
+	template <typename... Args>
+	static void print_debug(const Variant &p_arg1, const Args &...p_args) {
+		if (SentryOptions::get_singleton()->is_debug_enabled()) {
+			std::cout << "[sentry] DEBUG GODOTSDK ";
```
bruno-garcia: I'd leave out the 'godotsdk' part
```suggestion
			std::cout << "[sentry] DEBUG ";
```

src/sentry_util.h
```diff
@@ -1,16 +1,45 @@
...
+				std::cout << variant.stringify().utf8();
+			}
+			std::cout << std::endl;
+		}
+	}
+
+	template <typename... Args>
+	static void print_error(const Variant &p_arg1, const Args &...p_args) {
+		if (SentryOptions::get_singleton()->is_debug_enabled()) {
+			std::cerr << "[sentry] ERROR GODOTSDK ";
```
bruno-garcia:
```suggestion
			std::cerr << "[sentry] ERROR ";
```

project/project.godot
limbonaut: That's not ours, that's a Godot's option to enable file logging on platforms that usually have it disabled (like Android).
EDIT:
Our options are kept in `[sentry]` section.

src/sentry_util.h
limbonaut: Done

src/sentry_util.h
limbonaut: Done

PlasmaDev5 (approved):
Everything looks good to me.

project/project.godot
```diff
@@ -26,3 +30,4 @@ window/stretch/aspect="keep_height"
 
 config/dsn="https://3f1e095cf2e14598a0bd5b4ff324f712@o447951.ingest.us.sentry.io/6680910"
 config/debug=true
+config/error_logger/capture_type=1
```
bitsandfoxes, line 33: I take it this is the option that controls this feature? Why is it a `type`?

src/sentry_options.h
```diff
@@ -8,6 +8,12 @@
 #include <godot_cpp/variant/variant.hpp>
 
 class SentryOptions {
+public:
+	enum CaptureType {
```
bitsandfoxes, line 12: I'm not sure I understand the split between `capture as breadcrumb` and `capture as event`. Those are not mutually exclusive.

bitsandfoxes (approved):
This looks amazing! One thing just does not fully make sense to me:

The way I see this feature is that we're creating breadcrumbs out of logs and that errors that got logged to the logfile are automatically captured as events. After they have been captured they should get added as breadcrumbs as well. 

I mean we can iterate on it for sure so this is not a blocker but I think the functionality is a tiny bit off.

src/sentry_options.h
limbonaut: Makes sense! Thanks for reviewing. I added an issue to track it: #10

project/project.godot
limbonaut: You're right, it shouldn't be mutually exclusive.

Issue #10, opened 2024-11-04
Allow capturing errors as an event and a breadcrumb at the same time


limbonaut:
A previous error should show up as a breadcrumb in the next error.
Currently, we can choose one or the other, but not both.

PR #11, opened 2024-11-05
Internal SDK abstraction and project refactoring

limbonaut:
Most of the additions/deletions are simply due to restructuring of the code-base. Please, check the overview section below.

### Overview of changes
- Introducing namespaces
  - `sentry` namespace contains internals, interface to integrated SDKs & implementations, utility functions, etc.
  - Global namespace contains GDScript-exposed functionality, such as `SentrySDK` (entry point), and `SentryUser`
  - `sentry::contexts` -- factory functions for various Sentry contexts, e.g. `sentry::contexts::make_gpu_context()`
  - `sentry::native` for sentry-native specific code
  - `sentry::util` for utility functions, such as debug printing
- Class `Sentry` renamed => `SentrySDK` -- it's a user-facing entry point to sentry-godot API; exposed to GDScript.
  - Contains user-facing functions like `add_breadcrumb()` and `set_tag()`
  - Initializes and provides internal access to the proper `InternalSDK` instance (see below)
- New classes
  - `InternalSDK` -- interface to integrated SDKs; i.e. abstraction for SDKs that are used internally.
    - Contains declarations like `set_tag()` and `capture_message()`
    - Used internally, not exposed to GDScript
  - `NativeSDK` -- implements `InternalSDK` using sentry-native
  - `DisabledSDK` -- dummy class for when SDK is not enabled - all functions are noop
- Directory `src/sentry/` contains code for `sentry` namespace
- Directory `src/sentry/native` contains native-related code
- Global namespace code remains in `src/`


src/sentry_logger.cpp
```diff
@@ -118,41 +116,28 @@ void SentryLogger::_log_error(const char *p_func, const char *p_file, int p_line
 
 	if (as_breadcrumb) {
```
limbonaut, line 117: I'll address #10 in a separate PR

project/main.gd
```diff
@@ -24,51 +24,51 @@ func _ready() -> void:
...
-		Sentry.LEVEL_FATAL:
+		SentrySDK.LEVEL_FATAL:
 			level_choice.text = "FATAL"
 
 
 func _on_capture_button_pressed() -> void:
-	Sentry.capture_message(message_edit.text, _event_level)
-	print("Captured message event. Event ID: " + Sentry.get_last_event_id())
+	SentrySDK.capture_message(message_edit.text, _event_level)
+	print("Captured message event. Event ID: " + SentrySDK.get_last_event_id())
```
bruno-garcia: While this is a good example, the method capture_message should return the id already (at least if it successfully queued the event for submission). If the queue was overflowing, and capture_message already knows that event isn't going to be sent, it can return an empty GUID without `-` (SentryId is basically that, a GUID without `-`)

src/sentry/contexts.cpp
```diff
@@ -0,0 +1,340 @@
...
+			return "Portrait";
+		} break;
+		case DisplayServer::SCREEN_REVERSE_LANDSCAPE: {
+			return "Landscape (reverse)";
+		} break;
+		case DisplayServer::SCREEN_REVERSE_PORTRAIT: {
+			return "Portrait (reverse)";
+		} break;
+		case DisplayServer::SCREEN_SENSOR_LANDSCAPE: {
+			return "Landscape (defined by sensor)";
```
bruno-garcia, line 33: What does 'defined by sensor' mean here. In other words how does it compare to non-defined-by-sensor?

src/sentry/contexts.cpp
```diff
@@ -0,0 +1,340 @@
...
+
+	device_context["arch"] = Engine::get_singleton()->get_architecture_name();
+	int primary_screen = DisplayServer::get_singleton()->get_primary_screen();
+	String orientation = _screen_orientation_as_string(primary_screen);
+	if (orientation.length() > 0) {
+		device_context["orientation"] = orientation;
+	}
+
+	// TODO: Need platform-specific solutions - this doesn't work well.
+	String host = OS::get_singleton()->get_environment("HOST");
```
bruno-garcia, line 65: hostname can have someone's name (e.g: Bruno's MacBook). So we guard this behind a flag called `send_defautl_pii`

https://develop.sentry.dev/sdk/expected-features/data-handling/#sensitive-data

src/sentry/contexts.cpp
```diff
@@ -0,0 +1,340 @@
...
+	}
+	device_context["name"] = host;
+
+	String model = OS::get_singleton()->get_model_name();
+	if (!model.is_empty() && model != "GenericDevice") {
+		device_context["model"] = model;
+	}
+
+	Vector2i resolution = DisplayServer::get_singleton()->screen_get_size(primary_screen);
+	device_context["screen_width_pixels"] = resolution.x;
```
bruno-garcia, line 77: Nice, looks good:

https://docs.sentry.io/concepts/search/searchable-properties/issues/#devicescreen_width_pixels

src/sentry/contexts.cpp
```diff
@@ -0,0 +1,340 @@
...
+	device_context["screen_dpi"] = DisplayServer::get_singleton()->screen_get_dpi(
+			DisplayServer::get_singleton()->get_primary_screen());
+
+	Dictionary meminfo = OS::get_singleton()->get_memory_info();
+	// Note: Using double since int32 can't handle size in bytes.
+	device_context["memory_size"] = double(meminfo["physical"]);
+	device_context["free_memory"] = double(meminfo["free"]);
+	device_context["usable_memory"] = double(meminfo["available"]);
+
+	auto dir = DirAccess::open("user://");
```
bruno-garcia, line 88: Is this going to close only at the end of the scope?

We're doing a few more OS calls below which could be expensive (worth measuring) but I wonder if it's best to close these earlier.

src/sentry/contexts.cpp
```diff
@@ -0,0 +1,340 @@
...
+	device_context["processor_count"] = OS::get_singleton()->get_processor_count();
+	device_context["cpu_description"] = OS::get_singleton()->get_processor_name();
+
+	// Read/initialize device unique identifier.
+	String device_id = p_runtime_config->get_device_id();
+	if (device_id.length() == 0) {
+		device_id = sentry::uuid::make_uuid();
+		p_runtime_config->set_device_id(device_id);
+	}
+	device_context["device_unique_identifier"] = device_id;
```
bruno-garcia: Can we call it `installation_id` at least internally?
It's a concept understood internally that it won't allow identification of the user across apps, which is important for privacy:

```suggestion
	String installation_id = p_runtime_config->get_device_id();
	if (installation_id.length() == 0) {
		installation_id = sentry::uuid::make_uuid();
		p_runtime_config->set_device_id(installation_id);
	}
	device_context["device_unique_identifier"] = device_id;
```

https://docs.sentry.io/platforms/dotnet/guides/extensions-logging/enriching-events/identify-user/#id

https://docs.sentry.io/platforms/flutter/migration/#android-specific-changes

src/sentry/contexts.cpp
```diff
@@ -0,0 +1,340 @@
...
+
+Dictionary make_gpu_context() {
+	Dictionary gpu_context = Dictionary();
+	ERR_FAIL_NULL_V(RenderingServer::get_singleton(), gpu_context);
+	ERR_FAIL_NULL_V(OS::get_singleton(), gpu_context);
+
+	// Note: In headless/server mode, some of these functions return empty strings.
+	String adapter_name = RenderingServer::get_singleton()->get_video_adapter_name();
+	if (!adapter_name.is_empty()) {
+		gpu_context["name"] = RenderingServer::get_singleton()->get_video_adapter_name();
```
bruno-garcia:
```suggestion
		gpu_context["name"] = adapter_name;
```

src/sentry/contexts.cpp
```diff
@@ -0,0 +1,340 @@
...
+	ERR_FAIL_NULL_V(OS::get_singleton(), env_context);
+
+	env_context["sandboxed"] = OS::get_singleton()->is_sandboxed();
+	env_context["userfs_persistent"] = OS::get_singleton()->is_userfs_persistent();
+	env_context["granted_permissions"] = OS::get_singleton()->get_granted_permissions();
+
+#ifdef LINUX_ENABLED
+	String distribution_name = OS::get_singleton()->get_distribution_name();
+	if (!distribution_name.is_empty()) {
+		env_context["distribution_name"] = distribution_name;
```
bruno-garcia, line 266: Is this already part of the protocol?
If not, I wonder what we do with this already, perhaps the OS name could be this directly.

IIRC on Electron (@timfish can u confirm?) we looked up distro name too (don't think we do in .NET/Unity/Unreal, @bitsandfoxes can u confirm?).

src/sentry/contexts.cpp
```diff
@@ -0,0 +1,340 @@
...
+Dictionary make_performance_context() {
+	Dictionary perf_context = Dictionary();
+	ERR_FAIL_NULL_V(OS::get_singleton(), perf_context);
+	ERR_FAIL_NULL_V(Engine::get_singleton(), perf_context);
+	ERR_FAIL_NULL_V(Performance::get_singleton(), perf_context);
+	ERR_FAIL_NULL_V(RenderingServer::get_singleton(), perf_context);
+
+	if (!OS::get_singleton()) {
+		// Too early, bailing out.
+		return perf_context;
```
bruno-garcia: This is something that would be useful to know if debugging the SDK.

```suggestion
        if (options.debug) {
            Sentry::DiagnosticLogger::LogDebug("make_performance_context: no OS singleton available. Called too early?");
        }
		return perf_context;
```

Not sure where the logger would come from. In .NET for example we have an internal property on the options object which we use to pass the logger around. And the actual implementation can change depending on environment. By default it's a stdout thing, but on some frameworks without a terminal (desktop, web apps on Visual Studio/IIS) we can change the implementation. We also allow the user to override the implementation so they can dump our SDK logs to disk so they can share with us to troubleshoot stuff. And level is configurable.

More here:

https://docs.sentry.io/platforms/dotnet/configuration/diagnostic-logger/

src/sentry/contexts.cpp
```diff
@@ -0,0 +1,340 @@
...
+	}
+
+	// * Injecting "Performance" context...
+
+	// Static memory allocation.
+	perf_context["static_memory_peak_usage"] = String::humanize_size(OS::get_singleton()->get_static_memory_peak_usage());
+	perf_context["static_memory_usage"] = String::humanize_size(OS::get_singleton()->get_static_memory_usage());
```
bruno-garcia, line 284: I wonder if these could fit in the `Memory Info` context. Although I'm not sure we event made that officially part of the protocol.

Aligning these across SDKs has a few advantages like possibly creating some issue detectors based on this data in the future. We don't need to do this now, could be a ticket and follow up later but ideally we don't move on and forget about it.

src/sentry/contexts.cpp
```diff
@@ -0,0 +1,340 @@
...
+
+	Dictionary meminfo = OS::get_singleton()->get_memory_info();
+	perf_context["thread_stack_size"] = String::humanize_size(meminfo["stack"]);
```
bruno-garcia: Which thread stack is this about? The current one we're running?
This one seems to fit better in the `event.threads` part of the protocol, but that'd only make sense if `attachStacktrace` is set to true, so we'd walk at least the current thread's stack and attach, even on a `capture_message`

src/sentry/contexts.cpp
```diff
@@ -0,0 +1,340 @@
...
+		uint64_t buffer_mem = RenderingServer::get_singleton()->get_rendering_info(RenderingServer::RENDERING_INFO_BUFFER_MEM_USED);
+		perf_context["rendering_buffer_mem_used"] = String::humanize_size(buffer_mem);
+
+		// Frame statistics.
+		uint64_t draw_calls = RenderingServer::get_singleton()->get_rendering_info(RenderingServer::RENDERING_INFO_TOTAL_DRAW_CALLS_IN_FRAME);
+		perf_context["rendering_draw_calls"] = draw_calls;
+	}
+
+	// TODO: Collect more useful metrics: physics, navigation, audio...
+	// TODO: Q: Split into categories and make it optional?
```
bruno-garcia, line 316: Splitting might make sense.

But we can change these easily later so not a blocker

src/sentry/disabled_sdk.h
```diff
@@ -0,0 +1,33 @@
...
+
+	virtual void set_user(const Ref<SentryUser> &p_user) override {}
+	virtual void remove_user() override {}
+
+	virtual void add_breadcrumb(const String &p_message, const String &p_category, Level p_level,
+			const String &p_type = "default", const Dictionary &p_data = Dictionary()) override {}
+	// TODO: Consider adding the following function.
+	// virtual void clear_breadcrumbs() = 0;
+
+	virtual void capture_message(const String &p_message, Level p_level, const String &p_logger = "") override {}
```
bruno-garcia: e.g: https://github.com/getsentry/sentry-dotnet/blob/cb67d21ba8a9d1ecc23a853f6be2fc6e77a69316/src/Sentry/SentryId.cs


Helpful tip for building SDKs: If you take that id and paste it in Sentry's issue search, it'll open that specific event directly.

```suggestion
	virtual SentryId capture_message(const String &p_message, Level p_level, const String &p_logger = "") override {}
```

src/sentry/native/native_sdk.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+	// Establish handler path.
+	String handler_fn;
+#ifdef LINUX_ENABLED
+	handler_fn = "crashpad_handler";
+#elif WINDOWS_ENABLED
+	handler_fn = "crashpad_handler.exe";
+#elif MACOS_ENABLED
+	handler_fn = "crashpad_handler";
+#endif
```
bruno-garcia:
```suggestion
#ifdef LINUX_ENABLED || MACOS_ENABLED
	handler_fn = "crashpad_handler";
#elif WINDOWS_ENABLED
	handler_fn = "crashpad_handler.exe";
#else
    #error "Target not supported"
#endif
```

src/sentry/native/native_sdk.cpp
```diff
@@ -0,0 +1,207 @@
...
+		ERR_PRINT(vformat("Sentry: Failed to locate crash handler (crashpad) - backend disabled (%s)", handler_path));
+		sentry_options_set_backend(options, NULL);
+	}
+
+	sentry_options_set_database_path(options, (OS::get_singleton()->get_user_data_dir() + "/sentry").utf8());
+	sentry_options_set_sample_rate(options, SentryOptions::get_singleton()->get_sample_rate());
+	sentry_options_set_release(options, SentryOptions::get_singleton()->get_release());
+	sentry_options_set_debug(options, SentryOptions::get_singleton()->is_debug_enabled());
+	sentry_options_set_environment(options, sentry::environment::get_environment());
+	sentry_options_set_sdk_name(options, "sentry.native.godot");
```
bruno-garcia, line 189: We usually have a subset of events come through sentry.native so having it report the original version (of sentry-native) made sense.

In this, since all events of this SDK will be `sentry.native.godot` I suspect it's better if we override the version and have this SDK's version. So when we get a bug report, it's easy to know what folks are using. Otherwise we might make changes on this SDK withotu bumping sentry-native and it won't be possible for us to tell them apart.

Is `sentry_options_set_sdk_version` available? Otherwise we'll need to add it.

This would also apply to Unreal which wraps native on desktop, @tustanivsky how does it work there?

src/sentry/native/native_sdk.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+	// Attach LOG file.
+	// TODO: Decide whether log-file must be trimmed before send.
+	if (SentryOptions::get_singleton()->is_attach_log_enabled()) {
+		String log_path = ProjectSettings::get_singleton()->get_setting("debug/file_logging/log_path");
+		if (FileAccess::file_exists(log_path)) {
+			log_path = log_path.replace("user://", OS::get_singleton()->get_user_data_dir() + "/");
+			sentry_options_add_attachment(options, log_path.utf8());
+		} else {
+			WARN_PRINT("Sentry: Log file not found. Make sure \"debug/file_logging/enable_file_logging\" is turned ON in the Project Settings.");
```
bruno-garcia, line 200: Just saw below `print_debug` and `print_error`. Should we align on a single logging strategy here (ideally with configurable verbosity akin to diagnostic logger I mentioned above)?

src/sentry/uuid.cpp
```diff
@@ -0,0 +1,18 @@
...
+#endif
+
+namespace sentry::uuid {
+
+String make_uuid() {
+#ifdef NATIVE_SDK
+	return sentry::native::make_uuid();
+#else
+	// TODO: uuid for other platforms.
+	return "1234";
```
bruno-garcia: How about we just `#error` for now?

Better than surprising behavior

src/sentry_logger.cpp
```diff
@@ -118,41 +116,28 @@ void SentryLogger::_log_error(const char *p_func, const char *p_file, int p_line
...
-		const char *error_string = error_types[p_error_type];
-		sentry_value_set_by_key(data, "godot_error_type", sentry_value_new_string(error_string));
-
-		sentry_value_set_by_key(crumb, "data", data);
-		sentry_add_breadcrumb(crumb);
+		Dictionary data;
+		data["function"] = String(p_func);
+		data["file"] = String(p_file);
+		data["line"] = p_line;
+		data["godot_error_type"] = String(error_types[p_error_type]);
```
bruno-garcia, line 123: where are these error types maintained?

I'm thinking we could load them in the UI to give the user more context

src/sentry_logger.cpp
```diff
@@ -118,41 +116,28 @@ void SentryLogger::_log_error(const char *p_func, const char *p_file, int p_line
...
+
+		SentrySDK::get_singleton()->add_breadcrumb(
+				p_rationale,
+				"error",
+				p_error_type == ERROR_TYPE_WARNING ? sentry::LEVEL_WARNING : sentry::LEVEL_ERROR,
```
bruno-garcia: This is used also below, maybe a conversion function?

src/sentry_sdk.cpp
```diff
@@ -0,0 +1,139 @@
...
+		const String &p_type, const Dictionary &p_data) {
+	internal_sdk->add_breadcrumb(p_message, p_category, p_level, p_type, p_data);
+}
+
+String SentrySDK::get_last_event_id() const {
+	return internal_sdk->get_last_event_id();
+}
+
+void SentrySDK::set_tag(const String &p_key, const String &p_value) {
+	ERR_FAIL_COND_MSG(p_key.is_empty(), "Sentry: Can't set tag with an empty key.");
```
bruno-garcia, line 37: Is this a user failing failure? how does it manifest?

We have a rule to not have the SDK error/crash the users' app. No one likes their crash reporting service being the reason why their customers had issues. So what we try to do is rely heavily on the `option.debug` flag so that if that's set to true, anything going on with the SDK that could be useful for debugging the SDK is made available to the end user (e.g: console log).

So in this case, we'd log to console (if debug.true) and no-op this call (just return)

src/sentry_sdk.cpp
```diff
@@ -0,0 +1,139 @@
...
+	if (p_user->get_id().is_empty()) {
+		// Take user ID from the runtime config or generate a new one if it's empty.
+		String user_id = get_user()->get_id();
+		if (user_id.is_empty()) {
+			user_id = sentry::uuid::make_uuid();
+		}
+		p_user->set_id(user_id);
+	}
+
+	// Save user in a runtime conf-file.
```
bruno-garcia, line 59: Why would we do this? Not sure it's the right path. If the game crashes, we shouldn't be reporting the previous user's name on follow up events (other than that one crash itself of course)

src/sentry_sdk.cpp
```diff
@@ -0,0 +1,139 @@
...
+	// Load the runtime configuration from the user's data directory.
+	runtime_config.instantiate();
+	runtime_config->load_file(OS::get_singleton()->get_user_data_dir() + "/sentry.dat");
+
+#ifdef NATIVE_SDK
+	internal_sdk = std::make_shared<NativeSDK>();
+	enabled = true;
+#else
+	// Unsupported platform
+	internal_sdk = std::make_shared<DisabledSDK>();
```
bruno-garcia, line 120: would be nice to debug log here too: This is an unsupported platform. This is what you can expect: ...

bruno-garcia (approved):
Suggestions, questions and tips but no blockers.

This is looking good!

project/main.gd
limbonaut: I addressed it in the follow-up commit. AFAIK we don't have a Guid in Godot ecosystem. We currently use `String` to represent ID. Another option is to use a custom class - although it may be an overkill.

src/sentry/contexts.cpp
limbonaut: As per [DisplayServer doc](https://docs.godotengine.org/en/stable/classes/class_displayserver.html#enum-displayserver-screenorientation), `SCREEN_SENSOR_LANDSCAPE` can be either default or reversed, based on the sensor.

src/sentry/contexts.cpp
limbonaut: This part is broken right now, AFAIK this variable is not set by default, so it's `localhost` in my tests. I think, we need platform-specific solutions (#14).
As for PII, added an issue to track it as well: #13

src/sentry/contexts.cpp
limbonaut: I'll look into it while tackling issue #12. It is closed at the end of the scope, following RAII.

src/sentry/contexts.cpp
limbonaut: Thanks

src/sentry/contexts.cpp
limbonaut: Addressed in a commit.

src/sentry/contexts.cpp
limbonaut: There's `OS.distribution`, although sentry-native is not setting it in my case. Is it normal?

src/sentry/contexts.cpp
limbonaut: This part is not needed anymore, actually.

We currently don't have any custom logger. Just using the Godot's one. It's a good idea to `ERR_PRINT()` it (or use `ERR_FAIL_*` macros) - it ends up in the app's log, and also pushed to the **Errors** panel in the editor, so developers are made aware of the issue during test runs.

src/sentry/contexts.cpp
limbonaut: Are there docs for the `Memory Info` context? Couldn't find it in the usual place.

src/sentry/contexts.cpp
limbonaut: It's the size of the current thread stack in bytes ("humanized" to avoid int32 overflow) as reported by https://docs.godotengine.org/en/stable/classes/class_os.html#class-os-method-get-memory-info
Since it's run on the main thread, it is about the main thread.

src/sentry/native/native_sdk.cpp
limbonaut: Done

src/sentry/native/native_sdk.cpp
limbonaut: Knowing sentry-native version seems useful too. Is it viable to have both versions reported?

src/sentry/native/native_sdk.cpp
limbonaut: We can wrap it in a custom logger if we need one. This is a built-in godot macro. If we want this message to end up reported to the developer in the Godot's **Errors** panel - we should be using it.

`ERR_PRINT` and `WARN_PRINT` are standard idiomatic godot logger macros that wrap also a line, and a file from where this call is made. We can make our own logger too, but this should be taken into consideration, because users *expect* errors to show up in the **Errors** panel and in the app's log file.

src/sentry/uuid.cpp
limbonaut: Makes sense. Addressed.

src/sentry_logger.cpp
limbonaut: Here https://github.com/godotengine/godot/blob/04692d83cb8f61002f18ea1d954df8c558ee84f7/core/io/logger.h#L51-L56
As this is not a part of the public API yet, the code is the only documentation there is.
There are two PRs aiming to expose this AFAIK.

src/sentry_sdk.cpp
limbonaut: This is a no-op macro, so it aligns with what you described. Read it like this: If the condition is true, return from the function while also reporting it as an error via the godot logger with a custom message.
This macro would expand to something like this:
```cpp
if (__builtin_expect(!!(p_key.is_empty()), 0)) {
   godot::_err_print_error(__FUNTION__, ..., "Sentry: Can't set tag with an empty key.")
   return;
} else ...
```
There are several such macros: FAIL in this case means return from the function while registering it as an error. Godot has a policy to crash only on unrecoverable issues like some signals. There are also crash macros like  `CRASH_COND(predicate)` - so you'll know when you see such.

src/sentry_sdk.cpp
limbonaut: So the user_id is more like a session_id? Each time generated anew?

src/sentry_sdk.cpp
limbonaut: Done

src/sentry_logger.cpp
limbonaut: Done

limbonaut:
Thanks for your review and suggestions, @bruno-garcia.
I responded to your comments, some implemented in the follow-up commits.

project/main.gd
bruno-garcia: In .NET it's a `struct` but basically just has the Guid type as a backing field: https://github.com/getsentry/sentry-dotnet/blob/cb67d21ba8a9d1ecc23a853f6be2fc6e77a69316/src/Sentry/SentryId.cs#L20

It's fine to be a string here if you prefer, but it has be a guid represented as a string, without `-`s

src/sentry/contexts.cpp
bruno-garcia: Sorry I still don't understand the different between the two.

src/sentry/contexts.cpp
bruno-garcia: yeah the protocol has data that was introduced by dfiferent SDKs/platforms.

Sentry native is the most basic SDK we have, it's all C and is simple. That said, if it's safe to add, feel free to add that to sentry-native directly. This way crashes on Unreal, Unity and others can leverage the same data. This is true most if not all things we consider adding here that relies on OS API assuming those APIs are stable and pretty old.

src/sentry/contexts.cpp
bruno-garcia: No docs, I think that was just .NET adding and it got undocumented (even didn't use the right key convention which is snake case

src/sentry/contexts.cpp
bruno-garcia: Main worth adding `main_thread_stack_size` as a key then

src/sentry/native/native_sdk.cpp
bruno-garcia: We'll know the sentry-native version by knowing which Godot SDK version it is. Since the versions here are tagged here and dep version is fixed

bruno-garcia: Granted it's not as easy as just looking at the event in Sentry to figure out, but today we only have 1 version

src/sentry_sdk.cpp
bruno-garcia: usually user is set by the developer once a user logs in. For example.

On mobile we have an 'id' that's fixed (the installation id), we use that on release health so we can get a count of users affected.

So if this is that 'installation_id', it makes sense to reuse the same id.

Note that user has also email, and username. That's what I meant by not storing on disk and reloading. The id (installation_id) is probably fine although I'm not srue we do this on desktop (could I have a game where different users login?)

bruno-garcia: Lets figure out what to do here

src/sentry_sdk.cpp
bruno-garcia: Please mark "Resolve Conversation" so it's easier to know what requires my input

image

bruno-garcia:
Thank you!

bruno-garcia:
Feel free to merge when you're ready

project/main.gd
limbonaut: It is sourced by `sentry-native` for `NativeSDK` backend, so it should be alright.

src/sentry/contexts.cpp
limbonaut: Sometimes you can force an orientation, like lock the screen in the landscape mode. The difference is that there are two landscape orientations: default and reversed (up-side down?). And the "Landscape (defined by sensor)" here means the appropriate one is chosen automatically based on the sensors' data. At least, these are the enum values reported by the engine (or display server). Maybe we can use better string representations for these?

src/sentry_sdk.cpp
limbonaut: So currently we simply generate a random UUID and persist it for consistency’s sake. By default, that is the only thing that is generated and persisted. But if a developer is [setting the data manually](https://github.com/getsentry/sentry-godot/blob/main/src/sentry_user.h) (like email, username, and even ID) - that data is saved instead. The idea is that the dev pops up a custom form, gets the necessary bits and the user permission, sets it using the public API, and doesn't have to do it again or bother about its persistence. I am not really sure if this is the right approach, and games might have different requirements. But the thing is, if we would rely on dev calling `set_user` every time, it might come too late if an event shows up early in the app's lifecycle.

I was considering having this persistence optional, such as using `persist_user_data` boolean flag in the options. This would simply generate a fresh `user_id` each time the app is started, and not persist it. So it would behave like a session_id with that option off. What do you think? Just for clarity, we don't do IO constantly - the values are loaded from the file on start and cached in memory for later use.

src/sentry_sdk.h
```diff
@@ -0,0 +1,57 @@
...
+#include "sentry/level.h"
+
+#include <godot_cpp/core/object.hpp>
+#include <memory>
+
+using namespace godot;
+
+// Entry point for Sentry SDK users.
+// This singleton class exposes the public API of the Sentry SDK for the Godot Engine.
+class SentrySDK : public Object {
```
bitsandfoxes, line 15:
```suggestion
class SentrySdk : public Object {
```
I'm confused by the mixing of cases but I think we're going with PascalCase everywhere else?

src/sentry/contexts.cpp
```diff
@@ -0,0 +1,329 @@
...
+	Dictionary meminfo = OS::get_singleton()->get_memory_info();
+	perf_context["main_thread_stack_size"] = String::humanize_size(meminfo["stack"]);
+
+	double fps_metric = Engine::get_singleton()->get_frames_per_second();
+	if (fps_metric) {
+		perf_context["fps"] = fps_metric;
+	}
+
+	// Frames drawn since engine started - age metric.
+	perf_context["frames_drawn"] = Engine::get_singleton()->get_frames_drawn();
```
bitsandfoxes, line 295: Did you have something in mind with that metric? I wonder how useful it actually is. I. guess with the target fps we can make assumptions about the general performance?

src/sentry/contexts.cpp
```diff
@@ -0,0 +1,329 @@
...
+	// * Injecting "Performance" context...
+
+	// Static memory allocation.
+	perf_context["static_memory_peak_usage"] = String::humanize_size(OS::get_singleton()->get_static_memory_peak_usage());
+	perf_context["static_memory_usage"] = String::humanize_size(OS::get_singleton()->get_static_memory_usage());
+
+	Dictionary meminfo = OS::get_singleton()->get_memory_info();
+	perf_context["main_thread_stack_size"] = String::humanize_size(meminfo["stack"]);
+
+	double fps_metric = Engine::get_singleton()->get_frames_per_second();
```
bitsandfoxes, line 289: This is a fixed target value, right? Like the target fps?

src/sentry_sdk.h
limbonaut: @bitsandfoxes 
I went with this styling, because acronyms/initialisms are all capitalized in Godot with PascalCase, unlike C#/MSDN convention. For example:
- [ORMMaterial3D](https://docs.godotengine.org/en/stable/classes/class_ormmaterial3d.html)
- [JSON](https://docs.godotengine.org/en/stable/classes/class_json.html)
- [VisualShaderNodeScreenUVToSDF](https://docs.godotengine.org/en/stable/classes/class_visualshadernodescreenuvtosdf.html)
- [AudioStreamWAV](https://docs.godotengine.org/en/stable/classes/class_audiostreamwav.html)
- [SkeletonModification2DFABRIK](https://docs.godotengine.org/en/stable/classes/class_skeletonmodification2dfabrik.html)

src/sentry/contexts.cpp
limbonaut: Since we have an app start time, it's probably not very useful. It is an age metric in terms of frames drawn since the application started.

src/sentry/contexts.cpp
limbonaut: It's not target, it's average (aka FPS), sampled once per second.
https://github.com/godotengine/godot/blob/cb411fa960f0b7fdbd97dcdb4c90f9346360ee0e/main/main.cpp#L4478
https://github.com/godotengine/godot/blob/cb411fa960f0b7fdbd97dcdb4c90f9346360ee0e/core/config/engine.h#L122

src/sentry/contexts.cpp
limbonaut: > Did you have something in mind with that metric?

Yeah, I was considering removing it since we have app start time.

src/sentry/contexts.cpp
bruno-garcia: Yeah just wondering if it's worth adding the "sensor" one since it adds complexity. We could just map both to the same "Landscape", regardless if it's sensor one or not

limbonaut: @bruno-garcia 
I can make a PR. Technically, there's 3 different landscape values. What about "Landscape (reverese)"? Keeping or grouping it all under "Landscape"?

PR #15, opened 2024-11-12
Allow capturing errors as an event and a breadcrumb at the same time

limbonaut:
Implements #10

sentry-errors

### DIFF
--- a/src/sentry_logger.cpp
+++ b/src/sentry_logger.cpp
@@ -90,18 +90,24 @@ void SentryLogger::_process_log_file() {
 }
 
 void SentryLogger::_log_error(const char *p_func, const char *p_file, int p_line, const char *p_rationale, ErrorType p_error_type) {
-	bool as_breadcrumb = false;
+	bool as_breadcrumb = SentryOptions::get_singleton()->is_error_logger_capture_as_breadcrumb_enabled();
+	bool as_event = SentryOptions::get_singleton()->is_error_logger_capture_as_event_enabled();
+
 	if (p_error_type == ERROR_TYPE_WARNING) {
 		if (SentryOptions::get_singleton()->is_error_logger_log_warnings_enabled()) {
-			as_breadcrumb = true;
+			// Log warning only if capture is enabled and set to capture as a breadcrumb.
+			as_breadcrumb = true && (as_breadcrumb || as_event);
+			as_event = false;
 		} else {
 			// Don't log if warnings are disabled.
-			return;
+			as_breadcrumb = false;
+			as_event = false;
 		}
-	} else {
-		if (SentryOptions::get_singleton()->get_error_logger_capture_type() == SentryOptions::CAPTURE_AS_BREADCRUMB) {
-			as_breadcrumb = true;
-		}
+	}
+
+	if (!as_breadcrumb && !as_event) {
+		// Don't capture if capture is disabled.
+		return;
 	}
 
 	// Debug output.
@@ -114,22 +120,8 @@ void SentryLogger::_log_error(const char *p_func, const char *p_file, int p_line
 		printf("   Error Type: %s\n", error_types[p_error_type]);
 	}
 
-	if (as_breadcrumb) {
-		// Log error as breadcrumb.
-		Dictionary data;
-		data["function"] = String(p_func);
-		data["file"] = String(p_file);
-		data["line"] = p_line;
-		data["godot_error_type"] = String(error_types[p_error_type]);
-
-		SentrySDK::get_singleton()->add_breadcrumb(
-				p_rationale,
-				"error",
-				godot_error_to_sentry_level(p_error_type),
-				"error",
-				data);
-	} else {
-		// Capture error event.
+	// Capture error as event.
+	if (as_event) {
 		sentry::InternalSDK::StackFrame stack_frame{
 			.filename = p_file,
 			.function = p_func,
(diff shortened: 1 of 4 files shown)


limbonaut:
@bitsandfoxes
I reworked it a little, as we discussed.

image

src/sentry_logger.cpp
```diff
@@ -152,11 +134,27 @@ void SentryLogger::_log_error(const char *p_func, const char *p_file, int p_line
...
+
+	// Capture error as breadcrumb.
+	if (as_breadcrumb) {
+		Dictionary data;
+		data["function"] = String(p_func);
+		data["file"] = String(p_file);
+		data["line"] = p_line;
+		data["godot_error_type"] = String(error_types[int(p_error_type)]);
+
+		SentrySDK::get_singleton()->add_breadcrumb(
```
bruno-garcia, line 155: Based on the screenshot loks like the order is inverted (breadcrumb added before the error) but here it seems the breadcrumb is added after.

Do we first capture the error, and then add the breadcrumb so the next error contains a breadcrumb to the previous error?

src/sentry/godot_error_types.h
```diff
@@ -0,0 +1,39 @@
...
+
+// Enum used with bitwise operations to represent the set of Godot error types that the Sentry logger should capture.
+enum class GodotErrorMask {
+	NONE = 0,
+	ERROR = 1 << int(GodotErrorType::ERROR_TYPE_ERROR),
+	WARNING = 1 << int(GodotErrorType::ERROR_TYPE_WARNING),
+	SCRIPT = 1 << int(GodotErrorType::ERROR_TYPE_SCRIPT),
+	SHADER = 1 << int(GodotErrorType::ERROR_TYPE_SHADER),
+	ALL = ERROR | WARNING | SCRIPT | SHADER,
+	ALL_EXCEPT_WARNING = ERROR | SCRIPT | SHADER,
```
bruno-garcia, line 28: is `SCRIPT` very spammy? We'd want the default behavior to only capture events for something you really want to act on.

limbonaut: Any of those can be spammy. That largely depends on the project.

src/sentry_logger.cpp
limbonaut: > Do we first capture the error, and then add the breadcrumb so the next error contains a breadcrumb to the previous error?

Yes, as in code. I can swap order in options, if it's confusing.

limbonaut: Done

src/sentry/godot_error_types.h
bruno-garcia: In any regard you might want to make it less spammy as you can. So keeping just `ERROR` perhaps is the right call, instead of adding also the other two. 

I'm not familiar with this enough but if there are too many entries, it'll be noisy and incur overhead

limbonaut: To clarify things, `SCRIPT` are errors that occur in a GDScript code, and `ERROR` are for errors that occur in the native code (which could still originate in a GDScript call). 
Spammy errors is largely a matter of fixing these issues in the project. It can be spammy if errors are simply ignored by the developer. If it's a serious concern, perhaps keeping it disabled by default is the better choice, making it intentional if the dev wants it in their project.

bitsandfoxes: We've got the same issue in Unity with errors originating from within the `Update` loop. But we've got a "not so great" debouncer in place for people to opt-in if they so wish. That might be a nice new feature: https://github.com/getsentry/sentry-godot/issues/22

src/sentry_logger.cpp
```diff
@@ -24,15 +26,15 @@ const char *error_types[] = {
...
-	SentryLogger::ERROR_TYPE_ERROR,
-	SentryLogger::ERROR_TYPE_WARNING,
-	SentryLogger::ERROR_TYPE_SCRIPT,
-	SentryLogger::ERROR_TYPE_SHADER,
+GodotErrorType error_type_as_enum[] = {
+	GodotErrorType::ERROR_TYPE_ERROR,
+	GodotErrorType::ERROR_TYPE_WARNING,
+	GodotErrorType::ERROR_TYPE_SCRIPT,
+	GodotErrorType::ERROR_TYPE_SHADER,
+	GodotErrorType::ERROR_TYPE_ERROR,
```
bitsandfoxes, line 38: Are the duplicates intentional?

bitsandfoxes (approved):
If I read that correctly: We're capturing ALL errors out of the box. And we're creating breadcrumbs for ALL log types?

src/sentry_logger.cpp
limbonaut: Yes, those are the mappings for the array just a few lines above.
In essence, it means that we map:
- "ERROR" => ERROR_TYPE_ERROR
- "USER ERROR" => ERROR_TYPE_ERROR
...

limbonaut:
@bruno-garcia It seems everything is clear now. Can this be merged?

src/sentry/godot_error_types.h
limbonaut: I like the debouncer idea. We can throttle the misbehaving parts and avoid silencing the rest of the user's project.

bruno-garcia (approved):

src/sentry_logger.cpp
bitsandfoxes: I mean the individual enums as entries in array are duplicated.

limbonaut: Yes, and this is intentional. You need to look at both definitions at the same time:
```cpp
const char *error_types[] = {
	"ERROR",
	"WARNING",
	"SCRIPT ERROR",
	"SHADER ERROR",
	"USER ERROR",
	"USER WARNING",
	"USER SCRIPT ERROR",
	"USER SHADER ERROR"
};

GodotErrorType error_type_as_enum[] = {
	GodotErrorType::ERROR_TYPE_ERROR,
	GodotErrorType::ERROR_TYPE_WARNING,
	GodotErrorType::ERROR_TYPE_SCRIPT,
	GodotErrorType::ERROR_TYPE_SHADER,
	GodotErrorType::ERROR_TYPE_ERROR,
	GodotErrorType::ERROR_TYPE_WARNING,
	GodotErrorType::ERROR_TYPE_SCRIPT,
	GodotErrorType::ERROR_TYPE_SHADER,
};
```
The first array are the strings that we look for in the log file. And the second array are the error types we map to from the values of the first array. Notice that both arrays have the same number of elements. 
So `error_types[0] => error_type_as_enum[0];` and so on.
It's just a faster mapping than a hash map would be. I'll add comments to the code.

PR #19, opened 2024-11-15
GDScript-based unit testing

limbonaut:
Adding gdUnit4 unit testing framework to the demo project and a couple of test suites. The framework supports headless mode, and has a marketplace CI action. However, it can't be deployed as a submodule, unless we use symlink (it's not cross-platform then).

Review notes: First commit is the framework source (those 26K lines is this commit alone). 

EDIT: 
- Now importing as a **submodule**. See https://github.com/getsentry/sentry-godot/pull/19#issuecomment-2489607713 below.
- Creating a symlink for the testing framework during the build process on Linux and macOS.
  - Creating an NTFS junction on Windows due to elevated permissions required for symlinks.

image


limbonaut:
Can we use `git-subtree` as a potential alternative to submodule here?

bruno-garcia:
> Can we use `git-subtree` as a potential alternative to submodule here?

Doesn't look like we're using git submodule though, right? The files were added directly

> Review notes: First commit is the framework source (those 26K lines is this commit alone).

Can this be a submodule?

limbonaut:
> Can this be a submodule?

No, because we need a specific subfolder from that repo at a specific location in our repo.

bruno-garcia:
You suggested git-tree, is that an alternative?

It might be worth excluding development from Windows then, I dunno. Vendoring all this code seems quite suboptimal, it's 26kloc. And keeping things updated then with upstream will be painful this way

limbonaut:
> It might be worth excluding development from Windows then, I dunno. Vendoring all this code seems quite suboptimal, it's 26kloc.

Both GDScript frameworks have this disadvantage, maybe the other one is smaller LoC-wise. 

> And keeping things updated then with upstream will be painful this way

There is a built-in update routine, which checks for the update when you launch the Godot project. So it's a semi-automated process, and perhaps it handles the compatibility of the plugin version vs Godot version.

> You suggested git-tree, is that an alternative?

`git-subtree` is an alternative to `submodule`, and I'm wondering if it could be utilized here to better solve this issue. It's tailored for the same use-case as `submodule`, but much different in implementation - you keep the history locally. This history can be squashed, synced with upstream, so like a submodule that is in-tree. This command can also extract history of a specific subfolder from a repository, but I couldn't find a way to "import" history from a subfolder, which is what we need.

limbonaut:
## Changes
- Import gdUnit4 testing framework as a submodule
- Creating a symlink for the testing framework during the build process (on Linux and macOS)
  - Can also run `scons project/addons/gdUnit4` to create the symlink manually
- New PowerShell script to create the symlink on Windows
  - The script prompts for elevated privileges, as Windows requires them to create a symlink.
  - Note: This operation is not performed during the build process, as it requires elevated privileges, and prompting for them during the build would be confusing.

src/sentry_sdk.cpp
```diff
@@ -64,6 +64,7 @@ void SentrySDK::set_user(const Ref<SentryUser> &p_user) {
 
 void SentrySDK::remove_user() {
 	internal_sdk->remove_user();
+	runtime_config->set_user(memnew(SentryUser));
```
bitsandfoxes, line 67: Where is this coming from?

SConstruct
```diff
@@ -156,3 +159,27 @@ manifest = env.Substfile(
...
+        # Note: parameter `target` is a list of build targets.
+        assert(len(target) == 1)
+        assert(len(source) == 1)
+        dst = str(target[0])
+        src = str(source[0])
+        src = os.path.relpath(src, os.path.dirname(dst))
+        os.symlink(src, dst)
+        return 0
+
+    gdunit_symlink = env.Command(
```
bitsandfoxes: Can't we `env.command(powershell.exe...)` from here on Windows as well? I guess `Scons` will need to run with elevated privileges but there does not seem to be a way around this on windows anyway?

bitsandfoxes (approved):

bruno-garcia:
> Note: This operation is not performed during the build process, as it requires elevated privileges, and prompting for them during the build would be confusing.

How do we make it work in CI in this case?

SConstruct
bruno-garcia: Also, did you consider another approach? Like this: https://stackoverflow.com/questions/58038683/allow-mklink-for-a-non-admin-user

scripts/link-testing-framework.ps1
```diff
@@ -0,0 +1,39 @@
...
+    $args = "-NoProfile -ExecutionPolicy Bypass -File `"$PSCommandPath`""
+    Start-Process powershell -ArgumentList $args -Verb RunAs
+    exit
+}
+
+Write-Host "Running as Administrator..."
+Push-Location -Path $repoRoot
+
+# Create the symlink.
+New-Item -ItemType SymbolicLink -Path $dst -Target $src
```
bruno-garcia: Did you try `New-Item -Name <link> -ItemType Junction -Value <source>` ?

https://superuser.com/questions/1323598/create-symbolic-link-as-regular-user

src/sentry_sdk.cpp
limbonaut: Right, it's a quick fix for a bug I found while working on the unit tests and this PR.
Commit: c3ca779e6ebf2d082291f7f71e0598e8439400d7

SConstruct
limbonaut: @bruno-garcia 
Not sure which one of those you mean 😕 
Developer mode is not default, so failing a build without it seems unreasonable to me, at least for this.
With the hard link, I get "Access denied".
```
> mklink /H LinkedMusic Music
Access is denied.
```

EDIT:
https://github.com/getsentry/sentry-godot/pull/19#discussion_r1852236159 worked like a charm.

limbonaut:
> > Note: This operation is not performed during the build process, as it requires elevated privileges, and prompting for them during the build would be confusing.
> 
> How do we make it work in CI in this case?

Easy. We'll see if we have the privileges, if not: `cp -R` or `mv` should do the job.

scripts/link-testing-framework.ps1
limbonaut: > Did you try `New-Item -Name <link> -ItemType Junction -Value <source>` ?

Oh, this is great, it seems to work! Thanks!

.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "sentry-native"]
 	path = sentry-native
 	url = https://github.com/getsentry/sentry-native.git
+[submodule "modules/gdUnit4"]
+	path = modules/gdUnit4
+	url = https://github.com/MikeSchulze/gdUnit4.git
```
bruno-garcia, line 6: We can consider a fork in getsentry in the future

README.md
```diff
@@ -47,3 +47,5 @@ python -m pip install --upgrade scons
     ```
 
 In the Godot editor, you can adjust the Sentry SDK settings by going to `Project Settings -> Sentry -> Config`.
+
+If you are using Windows, run the `scripts\link-testing-framework.ps1` script. This will create a symlink to the gdUnit4 testing framework submodule in the demo project and prompt you for elevated privileges, which are required to create symlinks on Windows.
```
bruno-garcia: Is this still needed?

bruno-garcia (approved):
Just a note on the README.md, other than that, LGTM

README.md
limbonaut: Dropped.

PR #20, opened 2024-11-19
Update dynamic values in the Device context

limbonaut:
The current default is to set context values during initialization, which works for most cases. The only context that is currently updated is "godot_performance".

This PR introduces a code path that updates values in the standard `Device` context just before the event is sent.
Values that are dynamically updated in the Device context with this PR: `orientation`, `free_memory`, `usable_memory`, `free_storage`.

### DIFF
--- a/src/sentry/contexts.cpp
+++ b/src/sentry/contexts.cpp
@@ -73,6 +73,7 @@ Dictionary make_device_context(const Ref<RuntimeConfig> &p_runtime_config) {
 		device_context["model"] = model;
 	}
 
+	// Note: Godot Engine doesn't support changing screen resolution.
 	Vector2i resolution = DisplayServer::get_singleton()->screen_get_size(primary_screen);
 	device_context["screen_width_pixels"] = resolution.x;
 	device_context["screen_height_pixels"] = resolution.y;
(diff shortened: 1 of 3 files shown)


src/sentry/contexts.cpp
```diff
@@ -73,6 +73,7 @@ Dictionary make_device_context(const Ref<RuntimeConfig> &p_runtime_config) {
 		device_context["model"] = model;
 	}
 
+	// Note: Godot Engine doesn't support changing screen resolution.
```
bitsandfoxes, line 76: Like, ever? How do Godot games support different screens?

src/sentry/contexts.cpp
```diff
@@ -106,6 +107,24 @@ Dictionary make_device_context(const Ref<RuntimeConfig> &p_runtime_config) {
...
+	Dictionary device_context = Dictionary();
+	int primary_screen = DisplayServer::get_singleton()->get_primary_screen();
+	device_context["orientation"] = _screen_orientation_as_string(primary_screen);
+
+	const Dictionary &meminfo = OS::get_singleton()->get_memory_info();
+	// Note: Using double since int32 can't handle size in bytes.
+	device_context["free_memory"] = double(meminfo["free"]);
+	device_context["usable_memory"] = double(meminfo["available"]);
+
+	auto dir = DirAccess::open("user://");
```
bitsandfoxes, line 120: Do we need permissions for that?

bitsandfoxes (approved):

src/sentry/contexts.cpp
limbonaut: Godot doesn't change the physical resolution of your desktop environment, it changes resolution scale instead. I.e., projects 720p on 1080p resolution, for example, like consoles do. Most modern games do the same, some also support changing the real resolution. But this way it makes `Alt-Tab` faster, and fewer issues with it. Upscaling is cheap these days.

src/sentry/contexts.cpp
limbonaut: It's guaranteed to be writable, judging by the docs: https://docs.godotengine.org/en/stable/tutorials/io/data_paths.html#accessing-persistent-user-data-user

Also, some interesting bits:
> On mobile platforms, this path is unique to the project and is not accessible by other applications for security reasons.

> On HTML5 exports, user:// will refer to a virtual filesystem stored on the device via IndexedDB. (Interaction with the main filesystem can still be performed through the [JavaScriptBridge](https://docs.godotengine.org/en/stable/classes/class_javascriptbridge.html#class-javascriptbridge) singleton.)

src/sentry/contexts.cpp
bitsandfoxes: So the resolution reported is the actual, physical screen resolution and not the resolution the game was running in?

limbonaut: Yep. The rendering resolution is referred to as the viewport size (of the window), and you can have multiple windows as well, on different displays - it can get really complicated. You can also have nested viewports, so technically, the game could be rendered at higher resolution nested viewport, downscaled to window viewport size and upscaled to main screen resolution. :smile:  Not that I recommend it.
Reference to doc: https://docs.godotengine.org/en/stable/tutorials/rendering/multiple_resolutions.html

PR #28, opened 2024-11-25
Feature: Logger limits and throttling

### LINKED ISSUE #22: Add a `debouncer`
It’s easy for the SDK to end up capturing a ton of errors in a short amount of time. For example, if there's a script error in `_process`, the SDK can capture every single one. This could quickly burn through a user’s quota without providing much helpful info.

### Why this is a problem
- You can blow through your quota super fast i.e. with errors in `_process` or `_physics_process`
- Sending a bunch of errors in one go could impact performance, especially during gameplay.
- A bunch of duplicate errors are not going to be terribly useful anyway

### How this could help
Adding a debounce feature would make error reporting smarter by:

- Limiting frequency: Only capturing one error every X seconds, for example. We have a very naive implementation like this in the Unity SDK
- Skipping duplicates: If the same error happens repeatedly (like in a loop), don’t capture it every time. But how would we detect duplicates?

limbonaut:
- **Problems**
  - Repeatedly logging the same error in a loop
  - Errors recurring every frame
  - Multiple various errors overwhelming quotas
  - Specific error type spamming & overwhelming quotas (?)
- **Goals**
  - Protect the frame budget
  - Reduce impact on quotas
  - Increase coverage: avoid duplicates, and some errors not reported at all
- **Complex solution**
  - **Limit errors to N events reported per frame**
    - Rationale: Protect the frame budget.
    - Options: `limits/events_per_frame`, ~~`limits/breadcrumbs_per_frame`~~
  - **Limit errors to 1 error captured per (file, lineno) within a T interval.**
    - Rationale: Increase error coverage and reduce redundancy.
    - Options: `limits/repeated_error_window_ms`
  - **Limit to N events ~~& M breadcrumbs~~ reported within a T milliseconds interval**
    - Rationale: Reduce overall impact on quotas.
    - Options: `limits/throttle_events`, `limits/throttle_window_ms`, ~~`limits/throttle_breadcrumbs`~~
- **Previously implemented**
  - **Limiting the number of log lines read per frame, defaults to 100.**
    - Limitation: If a user prints something heavy at the start of a frame, some errors may be omitted later that same frame (unlikely scenario).
    - This will go away with the logger API landing in Godot.
  - **Enabling/disabling capture for specific error types (event/breadcrumb).**
    - Drawback: It is manual and reactive, requiring developers to create a new build with updated configuration.

Aiming to resolve #22

### DIFF
--- a/src/sentry_options.h
+++ b/src/sentry_options.h
@@ -22,13 +22,13 @@ public:
 		int parse_lines = 100;
 
 		// Protect frametime budget.
-		int events_per_frame = 1;
+		int events_per_frame = 5;
 
 		// Limit to 1 error captured per source line within T milliseconds window.
 		int repeated_error_window_ms = 1000;
 
 		// Limit to N events within T milliseconds window.
-		int throttle_events = 4;
+		int throttle_events = 20;
 		int throttle_window_ms = 10000;
 	};
 
--- a/src/sentry_options.h
+++ b/src/sentry_options.h
@@ -22,15 +22,13 @@ public:
 		int parse_lines = 100;
 
 		// Protect frametime budget.
-		int breadcrumbs_per_frame = 5;
 		int events_per_frame = 1;
 
 		// Limit to 1 error captured per source line within T milliseconds window.
 		int repeated_error_window_ms = 1000;
 
-		// Limit to N events and M breadcrumbs within T milliseconds window.
+		// Limit to N events within T milliseconds window.
 		int throttle_events = 4;
-		int throttle_breadcrumbs = 20;
 		int throttle_window_ms = 10000;
 	};
 
--- a/src/sentry_options.h
+++ b/src/sentry_options.h
@@ -16,10 +16,22 @@ class SentryOptions {
 
 public:
 	struct LoggerLimits {
-		int max_lines_parsed = 30;
-		int breadcrumbs_per_frame = 10;
-		int events_per_frame = 2;
-		int repeated_error_throttling_ms = 1000;
+		// TODO: Establish proper default limits with the Sentry team.
+
+		// Limit the number of lines that can be parsed per frame.
+		int parse_lines = 30;
+
+		// Protect frametime budget.
+		int breadcrumbs_per_frame = 5;
+		int events_per_frame = 1;
+
+		// Limit to 1 error captured per source line within T milliseconds window.
+		int repeated_error_window_ms = 1000;
+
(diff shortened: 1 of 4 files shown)


src/sentry_options.h
```diff
@@ -14,6 +14,26 @@ class SentryOptions {
...
+		int events_per_frame = 1;
+
+		// Limit to 1 error captured per source line within T milliseconds window.
+		int repeated_error_window_ms = 1000;
+
+		// Limit to N events and M breadcrumbs within T milliseconds window.
+		int throttle_events = 4;
+		int throttle_breadcrumbs = 20;
+		int throttle_window_ms = 10000;
+	};
```
limbonaut: @bruno-garcia @bitsandfoxes 
Looking for feedback on the defaults for these limits.

bitsandfoxes: We've got naive solution in the Unity SDK. We're throttling based on type. So only 1 error - does not matter what error - per second (default value).

bitsandfoxes (commented):
I think throttling breadcrumbs might be a bad idea. If I (as a user) were to look through the breadcrumbs I expect them to mirror the logs 1:1 and to be complete. 
I think it's fair to say: If it's too spammy - and too slow - then the user should log less?

Dropping events - like we do with sample rate - is fine because we're deduplicating. But dropping breadcrumbs sounds like a dangerous misrepresentation of "what happened" instead of helping fix stuff.

limbonaut:
> I think throttling breadcrumbs might be a bad idea. If I (as a user) were to look through the breadcrumbs I expect them to mirror the logs 1:1 and to be complete. I think it's fair to say: If it's too spammy - and too slow - then the user should log less?

Makes sense, although logging less may not always be an option, since errors are something you do not intentionally put into your source code, usually 😅

> Dropping events - like we do with sample rate - is fine because we're deduplicating. No real information is lost.. But dropping breadcrumbs sounds like a dangerous misrepresentation of "what happened" instead of helping fix stuff.

And what about "errors in a loop"? E.g. the same `(file, line)` error spamming 100+ times and driving other breadcrumbs out of the picture.

limbonaut:
@bitsandfoxes I've updated the PR, removing the breadcrumb limits. Some breadcrumbs can still be dropped for repeated errors, see `repeated_error_window_ms` and related code (starting `sentry_logger.cpp:L108`). 
Also https://github.com/getsentry/sentry-godot/pull/28#issuecomment-2520121564

bruno-garcia (approved):

PR #36, opened 2024-11-29
Fix crash on shutdown on Windows

limbonaut:
This PR fixes a crash on Windows related to memory release, adds sanity checks and debug prints.

### DIFF
--- a/src/register_types.cpp
+++ b/src/register_types.cpp
@@ -18,10 +18,9 @@ void _init_logger() {
 		return;
 	}
 	// Add experimental logger to scene tree.
-	SentryLogger *logger = memnew(SentryLogger);
 	SceneTree *sml = Object::cast_to<SceneTree>(Engine::get_singleton()->get_main_loop());
-	if (sml) {
-		sml->get_root()->add_child(logger);
+	if (sml && sml->get_root()) {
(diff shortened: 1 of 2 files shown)


src/register_types.cpp
```diff
@@ -55,9 +54,13 @@ void initialize_module(ModuleInitializationLevel p_level) {
 }
 
 void uninitialize_module(ModuleInitializationLevel p_level) {
-	if (p_level == MODULE_INITIALIZATION_LEVEL_CORE) {
-		memdelete(SentrySDK::get_singleton());
-		delete SentryOptions::get_singleton();
+	if (p_level == MODULE_INITIALIZATION_LEVEL_SCENE) {
```
bitsandfoxes, line 57: Is this the latest step? Could we uninstall even later?

bitsandfoxes (approved):

src/register_types.cpp
limbonaut: I'll see if I can push it to earlier/later stages. It's been problematic with LEVEL_CORE so far.

limbonaut: Tracking in https://github.com/getsentry/sentry-godot/issues/39

PR #37, opened 2024-11-29
GHA: Run unit tests in CI

limbonaut:
Run unit tests in CI using Windows and Linux runners. This will help us detect compilation and runtime issues early on.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,3 +20,8 @@ jobs:
   build-extension:
     name: 🔌 Build GDExtension
     uses: ./.github/workflows/build_gdextension.yml
+
+  unit-tests:
+    name: 🧪 Unit tests
+    needs: build-extension
+    uses: ./.github/workflows/unit_tests.yml
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -2,18 +2,6 @@ name: 🧪 Unit tests
 
 on:
   workflow_call:
-    inputs:
-      godot-version:
-        description: Godot Engine version to use for testing.
-        type: string
-        default: "4.3-stable"
-
-  workflow_dispatch:
-    inputs:
-      godot-version:
-        description: Godot Engine version to use for testing.
-        type: string
-        default: "4.3-stable"
 
 jobs:
   unit-tests:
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -3,7 +3,7 @@ name: ⚙️ CI
 on:
   push:
     branches:
-      - "main"
+      # - "main"
```
bruno-garcia:
```suggestion
      - "main"
```

.github/workflows/unit_tests.yml
```diff
@@ -0,0 +1,83 @@
...
+        include:
+          - runner: windows-latest
+            godot-suffix: win64.exe
+          - runner: ubuntu-latest
+            godot-suffix: linux.x86_64
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+        with:
+          submodules: false # don't initialize submodules automatically
```
bruno-garcia, line 22: Why the additional step instead of this?

limbonaut: Can't specify which submodules to check out... We don't need `sentry-native` and such, which also packs its own submodules on top. If you prefer simpler, I can remove that step.

.github/workflows/ci.yml
limbonaut: That was [temporary](https://github.com/getsentry/sentry-godot/pull/37#issuecomment-2520008953).
Dropped.

.github/workflows/unit_tests.yml
bruno-garcia: All good

limbonaut:
@bitsandfoxes @bruno-garcia
Good to merge? It's still waiting for approval.

bitsandfoxes (approved):
Awesome stuff!

.github/workflows/unit_tests.yml
```diff
@@ -0,0 +1,83 @@
...
+          scons project/addons/gdUnit4
+          chmod +x project/addons/sentrysdk/bin/crashpad_handler
+          echo "--- Rebuilding import cache.."
+          ${GODOT_BIN} --headless --editor --path project/ --quit-after 2000 || true
+          echo "--- Finished rebuilding import cache."
+
+      - name: Run tests
+        shell: bash
+        timeout-minutes: 5
+        run: |
```
bitsandfoxes, line 75: I assume this step fails if the tests fail?

limbonaut: Strangely, docs don't specify. I'll test.

limbonaut: ~~I'm not sure about `timeout-minutes` - docs don't specify if it fails the step. Need to check it.~~ (Works as expected).
According to docs, the process should return:
```
0   = all tests success
100 = ends with test failures
101 = ends with test warnings
```

PR #44, opened 2024-12-11
Semantic versioning

### LINKED ISSUE #43: Semantic versioning
For release process, we need version tracking. Ideally, we want to see where the build comes from (which commit) and differentiate release builds from others (custom builds, PR/CI builds, etc.). 

Godot currently doesn't track versions for extensions, but that might change. GDScript-based plugins usually have a version string following semantic versioning (not enforced). 

The SDK version should be reported in captured events.

limbonaut:
Version is stored in `SConstruct` file. On build, we generate `src/sdk_version.gen.h`. This is necessary in order to include commit SHA as  "+abcd012345" version metadata suffix. As such, this helps to ensure that even a custom build has an up-to-date version reported. The generated file is not versioned in Git. 

Closes #43


bruno-garcia (approved):
Nice and simple! thanks

Issue #47, opened 2024-12-20
User-facing event hooks


limbonaut:
We're missing user-facing callbacks for `before_send` and `before_crash`. 
Additionally, there is no user-facing representation of event objects, making it impossible to edit them. Converting `sentry_value_t` to a `Dictionary` and back seems too wasteful.

Depends on:
- #51

Issue #46, opened 2024-12-20
Integration tests for NativeSDK


limbonaut:
Implement integration tests for NativeSDK to cover areas not accessible through unit tests and to verify basic functionality of `sentry-native` integration.

limbonaut:
Partially implemented with isolated tests.

limbonaut:
Superseded by #339

PR #51, opened 2024-12-23
feat: Event objects

limbonaut:
This PR implements:
- `SentryEvent` objects, which can be used to create or modify an event, and a native implementation.
  - `NativeEvent` class -- implementation used with `NativeSDK`
- Add `SentrySDK.capture_event(event)`, `SentrySDK.create_event()` methods.
- Unit tests for the `SentryEvent` class and SDK methods.

In future PRs, we can implement accessing tags, breadcrumbs, contexts, fingerprint, etc.

Dependency for: 
- #47 
- #46

### DIFF
--- a/src/sentry/disabled_event.h
+++ b/src/sentry/disabled_event.h
@@ -14,6 +14,7 @@ private:
 	sentry::Level level = sentry::Level::LEVEL_INFO;
 	String release;
 	String dist;
+	String environment;
 
 protected:
 	static void _bind_methods() {}
@@ -40,6 +41,9 @@ public:
 
 	virtual void set_dist(const String &p_dist) override { dist = p_dist; }
 	virtual String get_dist() const override { return dist; }
+
+	virtual void set_environment(const String &p_environment) override { environment = p_environment; }
+	virtual String get_environment() const override { return environment; }
 };
 
 #endif // DISABLED_EVENT_H
--- a/src/sentry/disabled_event.h
+++ b/src/sentry/disabled_event.h
@@ -24,6 +24,8 @@ public:
 	virtual void set_timestamp(const String &p_timestamp) override { timestamp = p_timestamp; }
 	virtual String get_timestamp() const override { return timestamp; }
 
+	virtual String get_platform() const override { return ""; }
+
 	virtual void set_level(sentry::Level p_level) override { level = p_level; }
 	virtual sentry::Level get_level() const override { return level; }
 };
--- a/src/sentry/disabled_event.h
+++ b/src/sentry/disabled_event.h
@@ -12,6 +12,9 @@ private:
 	String timestamp;
 	sentry::Level level = sentry::Level::LEVEL_INFO;
 
+protected:
+	static void _bind_methods() {}
+
 public:
 	virtual String get_id() const override { return ""; }
 
--- a/src/sentry/disabled_event.h
+++ b/src/sentry/disabled_event.h
@@ -12,6 +12,7 @@ private:
 	String timestamp;
 	String logger;
 	sentry::Level level = sentry::Level::LEVEL_INFO;
+	String release;
 
 protected:
 	static void _bind_methods() {}
@@ -32,6 +33,9 @@ public:
 
 	virtual void set_logger(const String &p_logger) override { logger = p_logger; }
 	virtual String get_logger() const override { return logger; }
+
+	virtual void set_release(const String &p_release) override { release = p_release; }
+	virtual String get_release() const override { return release; }
 };
 
 #endif // DISABLED_EVENT_H
--- a/src/sentry/disabled_event.h
+++ b/src/sentry/disabled_event.h
@@ -13,6 +13,7 @@ private:
 	String logger;
 	sentry::Level level = sentry::Level::LEVEL_INFO;
 	String release;
+	String dist;
 
 protected:
 	static void _bind_methods() {}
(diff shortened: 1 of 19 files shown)


src/sentry_sdk.h
```diff
@@ -47,11 +54,17 @@ class SentrySDK : public Object {
 	Ref<SentryUser> get_user() const { return runtime_config->get_user(); }
 	void remove_user();
 
-	String capture_message(const String &p_message, sentry::Level p_level, const String &p_logger = "");
+	String capture_message(const String &p_message, sentry::Level p_level = sentry::LEVEL_INFO, const String &p_logger = "");
 	String get_last_event_id() const;
 
+	Ref<SentryEvent> create_event() const;
+	Ref<SentryEvent> create_message_event(const String &p_message, sentry::Level p_level = sentry::LEVEL_INFO, const String &p_logger = "");
```
limbonaut: Do we need to expose `logger`? It's unclear what it's used for.

src/sentry/disabled_event.h
```diff
@@ -0,0 +1,49 @@
+#ifndef DISABLED_EVENT_H
+#define DISABLED_EVENT_H
+
+#include "sentry_event.h"
+
+// Event class that is used when the SDK is disabled.
```
bitsandfoxes, line 6: Where does the need for a `DisabledEvent` come from? Can't we use the SentryEvent and still have the SDK disabled?

src/sentry/native/native_event.cpp
```diff
@@ -0,0 +1,114 @@
+#include "native_event.h"
+
+#include "sentry/level.h"
+#include "sentry/native/native_util.h"
+
+#include <sentry.h>
+
+namespace {
+
+inline void _sentry_value_set_or_remove_string_by_key(sentry_value_t value, const char *k, const String &v) {
```
bitsandfoxes, line 10: So we end up with shortcuts and instead of `event.remove_release()` we want users to use `event.set_release("")`?

src/sentry/native/native_sdk.cpp
```diff
@@ -188,6 +171,32 @@ String NativeSDK::capture_error(const String &p_type, const String &p_value, Lev
...
+}
+
+Ref<SentryEvent> NativeSDK::create_message_event(const String &p_message, sentry::Level p_level, const String &p_logger) {
+	sentry_value_t event_value = sentry_value_new_message_event(
+			native::level_to_native(p_level),
+			p_logger.utf8().get_data(),
+			p_message.utf8().get_data());
+	Ref<SentryEvent> event = memnew(NativeEvent(event_value));
+	return event;
+}
```
bitsandfoxes: I'm wondering, what's the usecase for this over `capture_message`?

src/sentry_sdk.h
bitsandfoxes: I don't see the use for logger either. Logging itself is setup during initialization, right?

bitsandfoxes (commented):
This looks good to me!
I'm just questioning the usecase of `create_message_event`.

src/sentry/disabled_event.h
limbonaut: `SentryEvent` is abstract and cannot be instantiated, and it's supposed to be implemented by `NativeEvent` class and such (for each internal SDK used). The reason behind `DisabledEvent` is to return a non-null value, so that it doesn't break scripts which may not be checking if `SentrySDK` is enabled. I could put this logic into `SentryEvent`, but it would add some bloated state to each event. I.e. `SentryEvent` is basically an interface.

src/sentry_sdk.h
limbonaut: Our error logger is set up at initialization and has nothing to do with this. I'm not even sure what this field is used for. This `logger` bit is a part of event payload:
- https://develop.sentry.dev/sdk/data-model/event-payloads/

Maybe we just remove this part from function call, but leave it as property of `SenteryEvent`?

src/sentry/native/native_event.cpp
limbonaut: Not sure if I understand the question. This is a helper method. It sets a string value on a `sentry_value` object, unless the string is empty, in which case it removes such key. Hence, the name of the method.
So we don't store such values:
`os.name = ""`
Instead, we remove the `os.name` key-value from the object.

src/sentry/native/native_sdk.cpp
limbonaut: Technically, this is not necessary, as you can add an event, set its message and level properties, and whichever other properties you need. So this can be removed - it's a shortcut for creating customized message events. I think I took it from some other SDKs, I'm not sure.

src/sentry/native/native_event.cpp
limbonaut: Oh, I think I see what you mean. We have a pair of setter/getter for each property of `SentryEvent`.
`event.release` is how we access it in GDScript. `event.release = "some_release"` is how we set it. So in order to remove the value, we would be doing `event.release = ""`

limbonaut: @bitsandfoxes Do you think we need explicit methods for removing event properties?

src/sentry_sdk.h
limbonaut: It's a part of sentry-native call, so I left it as is: 
https://github.com/getsentry/sentry-native/blob/3981ea7e7fd012524e4c04501c9dd647563aa391/include/sentry.h#L377-L380
We also have it in `capture_message` signature.

src/sentry/native/native_sdk.cpp
limbonaut: Removed in https://github.com/getsentry/sentry-godot/pull/51/commits/8e3ac2ce9645af78763cf817d9640e0a2accba52

Issue #48, opened 2024-12-20
`OptionConfiguration` - customize options via code


limbonaut:
Allow customizing options via code. One complication is that editing options with `sentry-native` is only possible before initialization, which might require delaying initialization since the GDScript module isn't available until later. Currently, we don't initialize early (see #39), but eventually, this will become a problem to solve.

References:
- https://github.com/getsentry/sentry-unity/pull/1888
- https://github.com/getsentry/sentry-unity/issues/1621

PR #56, opened 2025-01-07
feat: Improve & expose `SentryOptions` class

### LINKED ISSUE #55: Expose `SentryOptions` in the public API
We need to enable passing `SentryOptions` to a user-specified callback. The class is currently not exposed in the public API, and needs refactoring.

Dependency for:
- #47
- #48

limbonaut:
This PR improves `SentryOptions` and exposes it in the public API. This is needed for event hooks, in-code configuration support, integration tests and probably more.

Includes:
- Expose `SentryOptions` in the public API
- Add setters & bind properties
- New macros to streamline property binding in `ClassDB` (reduce boilerplate)
- Extract limits into its own public API class -- `SentryLoggerLimits`
- Refactor `SentryOptions` loading from project settings

Reviewer tip: it might be easier to look at changes commit-by-commit.

Resolves #55

Dependency for:
- #47
- #48

### DIFF
--- a/src/sentry/native/native_sdk.cpp
+++ b/src/sentry/native/native_sdk.cpp
@@ -193,7 +193,7 @@ void NativeSDK::initialize() {
 	ERR_FAIL_NULL(ProjectSettings::get_singleton());
 
 	sentry_options_t *options = sentry_options_new();
-	sentry_options_set_dsn(options, SentryOptions::get_singleton()->get_dsn());
+	sentry_options_set_dsn(options, SentryOptions::get_singleton()->get_dsn().utf8());
 
 	// Establish handler path.
 	String handler_fn;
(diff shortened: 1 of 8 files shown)


src/sentry/native/native_sdk.cpp
```diff
@@ -223,7 +223,7 @@ void NativeSDK::initialize() {
 
 	sentry_options_set_database_path(options, (OS::get_singleton()->get_user_data_dir() + "/sentry").utf8());
 	sentry_options_set_sample_rate(options, SentryOptions::get_singleton()->get_sample_rate());
-	sentry_options_set_release(options, SentryOptions::get_singleton()->get_release());
+	sentry_options_set_release(options, SentryOptions::get_singleton()->get_release().utf8());
 	sentry_options_set_debug(options, SentryOptions::get_singleton()->is_debug_enabled());
 	sentry_options_set_environment(options, sentry::environment::get_environment());
```
bitsandfoxes, line 228:
```suggestion
	sentry_options_set_environment(options, sentry::environment::get_environment()).utf8();
```
Do we need to do that here too?

bitsandfoxes (approved):
🔥

src/sentry/native/native_sdk.cpp
limbonaut: Yeah, as we discussed, this returns a C string handler which we need to pass to `sentry_native` call.
The data type has changed for `release` property to `godot::String`, so this is now necessary, or otherwise it won't compile.

PR #58, opened 2025-01-13
fix: Internal error reported on editor start

limbonaut:
This PR fixes internal error reported on editor start if sentry-native is not initialized (default for editor).
Error is in `SentryEvent.get_level()`.

Also, improve initialization code structure, so that `SentryNative` instance is not created if SDK is disabled.

Bug introduced by #51

#skip-changelog


src/sentry/native/native_event.cpp
```diff
@@ -62,6 +65,9 @@ void NativeEvent::set_level(sentry::Level p_level) {
 
 sentry::Level NativeEvent::get_level() const {
 	sentry_value_t value = sentry_value_get_by_key(native_event, "level");
+	if (sentry_value_is_null(value)) {
+		return sentry::Level::LEVEL_ERROR;
```
bitsandfoxes, line 72: Can we add a note that this is a workaround? This is trying to get a level from an event but not actually getting it.

limbonaut: :+1:  Explained in 8be101f969794e1c30ad353652ff59dc80efe111
It's not set by default, because level is an optional attribute.

Issue #39, opened 2024-12-06
Initialize SDK as early as possible


limbonaut:
Currently, we initialize at `MODULE_INITIALIZATION_LEVEL_SCENE`, which is not the earliest level. This is because core singletons are unavailable until this stage in the current stable Godot release.

A PR has already been merged that makes core singletons available at `INITIALIZATION_LEVEL_CORE`:

- godotengine/godot#98862](https://github.com/godotengine/godot/pull/98862

These changes are planned for version 4.4, which is expected to be released sometime next year. If we raise the minimum compatibility to 4.4, we’ll be able to initialize earlier. In the meantime, we might be able to use preprocessor checks as a workaround.

Related Godot issues:
- https://github.com/godotengine/godot-cpp/issues/1180
- https://github.com/godotengine/godot/issues/64975

limbonaut:
Since Godot 4.4 contains changes to singleton initialization, we can push the SDK to initialize even earlier, provided the user configuration script is not set. With the script, we need to wait for ScriptServer initialization anyway.

PR #60, opened 2025-01-14
feat: Event hooks & configuration via GDScript

### LINKED ISSUE #47: User-facing event hooks

### LINKED ISSUE #48: `OptionConfiguration` - customize options via code

limbonaut:
This PR introduces configuration via GDScript and adds the user-facing event hooks `before_send` and `on_crash`, as outlined in  https://develop.sentry.dev/sdk/expected-features/#before-send-hook and https://docs.sentry.io/platforms/native/configuration/filtering/. 

It introduces a new public API class called `SentryConfiguration` which users can extend in a script, and assign the script file in options to configure the SDK during initialization. However, due to the way scripting is initialized in the Godot Engine, this comes with a trade-off: a slightly later initialization. If a configuration script is assigned, initialization must be delayed until `ScriptServer` is ready to load and run the user script.

Resolves #47 
Resolves #48


src/sentry_sdk.cpp
```diff
@@ -135,28 +182,22 @@ SentrySDK::SentrySDK() {
 	}
 
 	if (enabled) {
-#ifdef NATIVE_SDK
```
limbonaut, line 138: FYI this code moved to _initialize()

src/sentry_sdk.cpp
```diff
@@ -135,28 +182,22 @@ SentrySDK::SentrySDK() {
...
-		// Unsupported platform
-		sentry::util::print_debug("This is an unsupported platform. Disabling Sentry SDK...");
-		enabled = false;
-#endif
-	}
-
-	if (!enabled) {
-		sentry::util::print_debug("Sentry SDK is DISABLED! Operations with Sentry SDK will result in no-ops.");
-		internal_sdk = std::make_shared<DisabledSDK>();
-		return;
```
limbonaut, line 150: FYI this code moved to _intialize()

project/example_configuration.gd
```diff
@@ -0,0 +1,26 @@
...
+## Example Sentry configuration script.
+##
+## Tip: You can assign configuration script in the project settings.
+
+
+## Configure Sentry SDK options
+func _configure(options: SentryOptions) -> void:
+	print("[example_configuration.gd] Configuring SDK options via GDScript")
+	options.debug = true
+
```
bitsandfoxes, line 13: We could extend the sample here and provide some more options. I.e. setting the release and/or the environment. To give an idea what's there.

project/example_configuration.gd
```diff
@@ -0,0 +1,26 @@
...
+	# Set up event callbacks
+	options.before_send = _before_send
+	options.on_crash = _on_crash
+
+
+## before_send callback example
+func _before_send(ev: SentryEvent) -> SentryEvent:
+	print("[example_configuration.gd] Processing event: ", ev.id)
```
bitsandfoxes, line 21: One of the more important uses of this callback is to allow users filtering of events. I.e. `return null` for dropping events entirely. Based on the native implementation it looks like this should be supported. Can we update the sample with some filtering?

src/sentry/native/native_event.cpp
```diff
@@ -111,13 +111,20 @@ String NativeEvent::get_environment() const {
...
 NativeEvent::NativeEvent(sentry_value_t p_native_event) {
-	native_event = p_native_event;
+	if (sentry_value_refcount(p_native_event) > 0) {
+		sentry_value_incref(p_native_event); // acquire ownership
+		native_event = p_native_event;
+	} else {
+		// Shouldn't happen in healthy code.
+		native_event = sentry_value_new_event();
+		ERR_PRINT("Sentry: Internal error: Event refcount is zero.");
+	}
```
bitsandfoxes, line 121: Where is this change coming from?

src/sentry_sdk.cpp
```diff
@@ -91,6 +92,48 @@ void SentrySDK::_init_contexts() {
...
+		internal_sdk = std::make_shared<NativeSDK>();
+#else
+		// Unsupported platform
+		sentry::util::print_debug("This is an unsupported platform. Disabling Sentry SDK...");
+		enabled = false;
+#endif
+	}
+
+	if (!enabled) {
+		sentry::util::print_debug("Sentry SDK is DISABLED! Operations with Sentry SDK will result in no-ops.");
```
bitsandfoxes, line 109: I'm not quite following the whole initialization. Is logging set up prior to initialization? Does this show up 100% of the time with the SDK disabled? The way to get rid of this would be to increase the logging verbosity I guess?

bitsandfoxes (commented):
This is amazing stuff. The changelog reads as if the SDK initializes at different points depending on whether the configure callback has been implemented. I don't quite follow the chain of notifications to get from `init` to `configure` back to `actual init` with the actual options. Would unifying the initialization to just the one right after the script server runs simplify this?

project/example_configuration.gd
limbonaut: We can, although, this would apply to everything in the demo project, including our unit tests, so it should be something harmless.

src/sentry/native/native_event.cpp
limbonaut: I adjusted how ownership is handled in the `NativeEvent` class, just so we don't have to preclude constructor call with incref every time.

project/example_configuration.gd
limbonaut: I like the idea, although, we can't change environment until #66 is merged. Will keep it in mind.
I've updated the example configuration with setting release attribute in 46c122e68a1a58aee65477ec761adbea4c912433

src/sentry_sdk.cpp
limbonaut: If you are commenting on this specific if-branch, then this message would appear in stdout if `options.debug` is ON and SDK is disabled either manually via options or because it is unsupported platform. 

I'm not sure what you mean by logging here. Godot doesn't employ the logger concept, with verbosity levels and such. There is print, push_error, and push_warning in GDScript. But we're not adding this message to such channels, and similar to sentry_native, it is simply printed to stdout (for consistency’s sake). Our error logger will eventually process errors and warnings, but will not deal with prints.

project/example_configuration.gd
limbonaut: Updated the example with filtering in 9a49f11e9c0cf47f54cf41cde608ccb6cc8da91b

limbonaut:
Thanks for reviewing @bitsandfoxes. It's a best-effort solution among very few options. Let me explain. 

> The changelog reads as if the SDK initializes at different points depending on whether the configure callback has been implemented. 

Correct, more on that below.

> Would unifying the initialization to just the one right after the script server runs simplify this?

We don't have an initialization opportunity at that point. Here are our options:
1. Before the `ScriptServer` is initialized -- no user configuration is possible. Several opportunities to hook up here. We use this method if the user configuration script is not supplied.
2. Via autoload singleton, for which we need a user script. If we don't have a script - we can't use this option. This is our current approach if the user provides a script. Otherwise, we default to early initialization, and potentially very early initialization when #39 gets implemented.
3. Initialize late, during the main loop's first iteration - that's kinda late for us. Hence, I went with 1 and 2 combined.

These are the options I could find to tap into initialization, and I don't see any other way aside from introducing upstream changes.

> I don't quite follow the chain of notifications to get from `init` to `configure` back to `actual init` with the actual options. 

1. We're registering an autoload singleton (aka user script that extends our `SentryConfiguration` class) in the project settings. Only a script can be registered as an autoload singleton. This step takes place in the `SentrySDK` constructor. 
2. Autoload singleton script is loaded and added to the scene tree by the engine.
3. The `SentryConfiguration` instance (user script must extend this class) receives `NOTIFICATION_READY` when added to the scene tree. In C++, our notification processing code calls `_configure()` on the user script and then calls `notify_options_configured()` in SentrySDK.
4. `_notify_options_configured()` finally initializes the internal SDK. 
5. We also have a safety catch (implemented via option 3), just in case the user script fails for whatever reason, to notify the user about the script problem and initialize anyway at that point.

bitsandfoxes (approved):

PR #73, opened 2025-01-27
feat: Add auto debug mode

### LINKED ISSUE #67: Implement Auto Debug Mode
As recommended in [SDK Features](https://develop.sentry.dev/sdk/expected-features/#auto-debug-mode).

~Make debug option an enum: AUTO, ENABLED, DISABLED in the project settings and `SentryOptions`.~

limbonaut:
Add auto debug mode as outlined in [SDK Features](https://develop.sentry.dev/sdk/expected-features/#auto-debug-mode).

Now the option looks like this:
image

And in code:
```gdscript
options.debug = true
```

Closes #67


limbonaut:
Refactored PR to preserve the `debug` option as a boolean in code.

src/sentry_options.cpp
```diff
@@ -54,7 +55,7 @@ void SentryOptions::_define_project_settings(const Ref<SentryOptions> &p_options
 	_define_setting("sentry/config/dsn", p_options->dsn);
 	_define_setting("sentry/config/release", p_options->release);
 	_define_setting("sentry/config/dist", p_options->dist);
-	_define_setting("sentry/config/debug", p_options->debug);
+	_define_setting(PropertyInfo(Variant::INT, "sentry/config/debug", PROPERTY_HINT_ENUM, "Off,On,Auto"), (int)SentryOptions::DEBUG_DEFAULT);
```
bruno-garcia: In the UI it would be nice to expand on what this means somehow. 

is property hint menu that? Just by name folks won't know this is a local debug log of sentry itself as it can seem like "yeah I want sentry for debugging"

limbonaut: Good point, however, we don't have such a level of freedom with project settings. Even tooltips are not currently available. See https://github.com/godotengine/godot-proposals/issues/11294. We may consider naming the property more expressively, or the menu options.

The `PROPERTY_HINT_*` is a hint for the editor on how to edit this value - in this case, we say it's an enum with specific values, so the editor presents a drop-down menu with a choice.

bruno-garcia (approved):

src/sentry_options.cpp
limbonaut: @bruno-garcia 
As we established, I renamed the setting to "Debug Printing" in the Project Settings (b44e7cb78b2a28fb722afc424aa12e84a7551387). This is much clearer in this context, and we still retain standard `options.debug` which is a boolean. The options don't need to correspond to Project Settings properties, that is mostly for consistency. In this case, it's a good deviation.

If we value tooltips, I can also look into making a PR upstream to get that functionality exposed for extensions. Realistically, the earliest it can be merged is Godot 4.5 which I estimate to come out roughly in a year. The 4.4 is in the feature freeze now.

PR #74, opened 2025-01-29
feat: Add `dist` property to `SentryOptions`

limbonaut:
We have the `dist` property in `SentryEvent`, but it's missing from `SentryOptions`.

Also, remove the historical `String()` casts.


limbonaut:
Thanks for reviewing! 🙂
Just in time for #75

PR #124, opened 2025-02-22
docs: Add CONTRIBUTING & update README

limbonaut:
#skip-changelog as it's just repo docs

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,86 @@
+# Contributing Guide
+
+## Building SDK
+
+Sentry SDK for Godot Engine can be built for Windows x86_64, Linux x86_64, and macOS universal platforms. Support for more platforms and architectures are expected to be added in time.
+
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,5 @@
+repos:
+  - repo: https://github.com/pre-commit/mirrors-clang-format
+    rev: v19.1.7
```
bruno-garcia: can we use commit sha instead? The version could be in a comment:
```suggestion
    rev: 1923012abc38abc12 # v19.1.7
```

This way it avoid security risk of someone redefining that tag onto another commit if the repo is compromised

limbonaut: Committed in 5aafc8ad0b3b5db15aa4e1bb4c215f268a4d9e30 + c7ab177f0606becb63e048737ba8c50c5933fe53

PR #128, opened 2025-02-26
feat: Capture screenshots

### LINKED ISSUE #49: Capture screenshots
Investigate possibility for capturing screenshots on events. Grabbing the default viewport texture might be enough.

limbonaut:
This PR adds a new feature for attaching screenshots. 

Add a new option `attach_screenshot` which is disabled by default.

image

Resolves #49

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Features
 
 - In-editor class reference documentation ([#104](https://github.com/getsentry/sentry-godot/pull/104))
+- Capture screenshots ([#128](https://github.com/getsentry/sentry-godot/pull/128))
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 ### Features
 
 - In-editor class reference documentation ([#104](https://github.com/getsentry/sentry-godot/pull/104))
-- Capture screenshots ([#128](https://github.com/getsentry/sentry-godot/pull/128))
+- Capture screenshots when enabled via `attach_screenshot` option ([#128](https://github.com/getsentry/sentry-godot/pull/128))
 
 ### Fixes
 
--- a/src/sentry/native/native_sdk.cpp
+++ b/src/sentry/native/native_sdk.cpp
@@ -6,11 +6,14 @@
 #include "sentry/native/native_event.h"
 #include "sentry/native/native_util.h"
 #include "sentry/util.h"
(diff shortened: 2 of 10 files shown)


src/sentry/native/native_sdk.cpp
```diff
@@ -52,9 +71,10 @@ inline void inject_contexts(sentry_value_t p_event) {
 	}
 }
 
-sentry_value_t handle_before_send(sentry_value_t event, void *hint, void *closure) {
+sentry_value_t _handle_before_send(sentry_value_t event, void *hint, void *closure) {
 	sentry::util::print_debug("handling before_send");
-	inject_contexts(event);
+	_save_screenshot();
```
limbonaut, line 76: Unsure if we should attach screenshots on non-crash events like "message" events.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ### Features
 
 - In-editor class reference documentation ([#104](https://github.com/getsentry/sentry-godot/pull/104))
+- Capture screenshots ([#128](https://github.com/getsentry/sentry-godot/pull/128))
```
bruno-garcia: We should make this opt-in. But we can add this to the install snippets, so its' easier to show that this exists while making it obvious how to 'turn it off'.

```suggestion
- Capture screenshots when enabled via `attach_screenshot` option ([#128](https://github.com/getsentry/sentry-godot/pull/128))
```

src/sentry/native/native_sdk.cpp
bruno-garcia: on mobile we only added it on crashes. Would be nice to make this configurable actually. So by default it could be on all events (error+message) but one could configure to be only on errors.

CHANGELOG.md
limbonaut: It's currently an option flag that is turned off by default, similar to other SDKs.

Suggestion committed in 33332a5fc9e365b1152e64dfb240a1dab6ea1ef2

src/sentry/native/native_sdk.cpp
limbonaut: We don't differentiate internally yet, but technically we have messages (aka manual events), Godot errors, and crashes.

bruno-garcia (approved):

src/sentry/native/native_sdk.cpp
limbonaut: We can improve on this in a follow-up. Tracked in #131.

PR #135, opened 2025-03-05
feat: Show debug messages in Godot Output panel

limbonaut:
- Reroute local and sentry-native debug output to Godot Output panel and log file.
- Add `debug_verbosity` option to control the minimum message level that should be printed.
- Add verbosity controls in the Demo project, defaulting to INFO level.

image

image


src/sentry/util/print.h
```diff
@@ -0,0 +1,69 @@
...
+	if (!SentryOptions::get_singleton()->is_debug_enabled() && p_level < sentry::LEVEL_ERROR) {
+		return;
+	}
+	if (SentryOptions::get_singleton()->get_debug_verbosity() > p_level) {
+		return;
+	}
+
+	switch (p_level) {
+		case LEVEL_DEBUG:
+			UtilityFunctions::print("DEBUG: Sentry: ", p_arg1, p_args...);
```
bruno-garcia, line 25: How about we start with Sentry: then the level? So it's more consistent the 'source' of the logger coming first?

bruno-garcia (approved):

src/sentry/util/print.h
limbonaut: I've reordered this to ensure log-file consistency. Errors in the log file always start with `ERROR: ...` or `WARNING: ...`. When an error is generated using `ERR_FAIL_X` macros, it appears as `ERROR: Sentry: Something...`, meaning we cannot change this order without causing inconsistency. To maintain uniformity in the Godot log file, I’ve adjusted this output accordingly, as it is also logged.

PR #137, opened 2025-03-10
ci: Upload symbols via symbol-collector

limbonaut:
This PR adds steps to craft configuration to upload symbols for our builds via symbol-collector.

### DIFF
--- a/.craft.yml
+++ b/.craft.yml
@@ -6,3 +6,15 @@ targets:
   - name: registry
     sdks:
       github:getsentry/sentry-godot:
+  - name: symbol-collector
+    includeNames: /addons/sentrysdk/bin/macos/libsentrysdk\.macos\.(release|debug)\.framework/libsentrysdk\.macos\.(release|debug)
+    batchType: macos
+    bundleIdPrefix: sentry-godot-macos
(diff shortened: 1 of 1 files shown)


.craft.yml
```diff
@@ -6,3 +6,15 @@ targets:
   - name: registry
     sdks:
       github:getsentry/sentry-godot:
+  - name: symbol-collector
+    includeNames: /addons/sentrysdk/bin/macos/libsentrysdk\.macos\.(release|debug)\.framework/libsentrysdk\.macos\.(release|debug)
+    batchType: macos
+    bundleIdPrefix: sentry-godot-macos
+  - name: symbol-collector
```
limbonaut, line 13: I assumed here, that we can declare several `symbol-collector` steps.

limbonaut:
@bruno-garcia Do we need anything else to have release symbols collected?

bruno-garcia:
We don't have a "Sentry bucket" for Windows macOS or Linux symbols.
@mitsuhiko is this something we should consider adding instead of having every project upload our debug symbols for all SDKs that provide precompiled libraries?

Issue #110, opened 2025-02-09
Fix crashpad_handler missing executable permissions in exported projects on Linux


limbonaut:
For some reason this issue is Linux-specific. On macOS, crashpad_handler has proper executable permissions assigned. This may be a Godot Engine bug.

monitz87:
I'm actually experiencing crashes on macOS release builds due to `crashpad_handler` not having executable permissions

limbonaut:
@monitz87 Could you tell me which version you’re using, both addon and Godot? Also, if there are any steps I can take to try and reproduce this issue, I’d really appreciate it. I've just ran a bunch of exports using Godot 4.4 and all resulted in executable bits present on the exports. However, I'm actually testing this on macOS. I also tried removing the executable bits manually, and the worst that happened was that the SDK couldn’t initialize, but the app kept running and didn’t crash. So I’m not sure what’s actually going wrong in your case that causes the crash. If you could share a stack trace from a terminal app, that would be super helpful!

monitz87:
What crashes is not the game itself, but the parallel process that runs `crashpad_handler`.

I'm using godot 4.4, macOS version 15.3.1 (Sequoia), sentry SDK version 0.3.0


We're also experiencing stutters in the game whenever sentry reports an error to the server. Should I open a separate issue for that?

limbonaut:
@monitz87  Thanks for the info! I’m still not sure why the `crashpad_handler` loses its executable bit. We try to fix it when the editor starts, so even if it’s unpacked in a way that it loses the executable permission, it gets fixed when the editor restarts. And exporting a project in my tests preserves it. So I’m not sure how it’s lost. I'll look into applying such a fix at runtime, if it's feasible.
 
> We're also experiencing stutters in the game whenever sentry reports an error to the server. Should I open a separate issue for that?

I’d like to take a closer look at this. Let's create a new issue for it. And can you also check if any of the following removes the stutters:
- Disable the `debug` option in the project settings.
- Disable the `attach_screenshot` option in the project settings.

I'll profile the code on my side in the meantime. And if you can provide a log file from one of the stutter-triggering runs, that may also help.

monitz87:
Thanks a lot! Once I get a reliable repro I'll create a new issue and try out your suggestions

PR #148, opened 2025-03-21
ref: Optimize screenshot feature to reduce stutters

limbonaut:
Note: Saving image from the root viewport is a slow operation: ~20 ms for JPG, 40 ms for PNG.

* ~~Don't take screenshots on non-crash events as it's too expensive.~~
* Use JPG format to speed up compression.
* Ensure that a screenshot is taken no more than once per frame drawn.
* Mention performance impact in the built-in documentation.

Addressing https://github.com/getsentry/sentry-godot/issues/110#issuecomment-2743714520

CI fails because of #147.

### Tasks
- [ ] Mention in docs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixes
 
-- Disable screenshot for non-crash events to address logger stutters ([#148](https://github.com/getsentry/sentry-godot/pull/148))
+- Optimize screenshot feature to reduce stutters ([#148](https://github.com/getsentry/sentry-godot/pull/148))
 - Don't process screenshot when `attach_screenshot` option is disabled ([#145](https://github.com/getsentry/sentry-godot/pull/145))
 
 ### Dependencies
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- Disable screenshot for non-crash events to address logger stutters ([#148](https://github.com/getsentry/sentry-godot/pull/148))
+
 ### Dependencies
 
 - Bump Native SDK from v0.8.1 to v0.8.2 ([#144](https://github.com/getsentry/sentry-godot/pull/144))
--- a/doc_classes/SentryOptions.xml
+++ b/doc_classes/SentryOptions.xml
@@ -14,7 +14,7 @@
 			If [code]true[/code], the SDK will attach the Godot log file to the event.
 		</member>
 		<member name="attach_screenshot" type="bool" setter="set_attach_screenshot" getter="is_attach_screenshot_enabled" default="false">
-			If [code]true[/code], the SDK will try to capture and attach a screenshot to the event.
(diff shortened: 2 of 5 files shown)


limbonaut:
> Am I right to assume the screenshot was still being taken but simply not sent with the event?

@monitz87 Yes, that's exactly what is happening. While attachment is properly disabled, the processing is still running. And the other PR should fix that oversight. I might look into having faster screenshots later. Grabbing image from the root viewport is sadly too slow.

monitz87:
Thanks for fixing it 😀 

I'll be waiting for the new release (tried to build it myself but couldn't do it for windows)

bitsandfoxes (commented):
I'm not convinced we should drop screenshots on non-crash events. There will always be some overhead when capturing. Going from 40ms to 20 is already a huge gain.

limbonaut:
@bitsandfoxes  20ms is still a lot, meaning that it practically guarantees stutter on errors in any game that targets 60 FPS (16ms frame budget). And script errors are not that uncommon. What if we make screenshots on non-crash events optional (opt-in or opt-out)?

monitz87:
My 2 cents: 20ms is unacceptably long for any game aiming for a decent framerate. Non-critical errors are very common and sometimes unavoidable (godot itself has a bunch of yet unsolved bugs that cause errors during runtime). At least for our use case, a 20ms stutter every time there is an error makes this library a non-starter

(I don't mean to sound demanding, we just really really need sentry, but we won't be able to use it with that limitation)

RafaelVidaurre:
> I'm not convinced we should drop screenshots on non-crash events. There will always be some overhead when capturing. Going from 40ms to 20 is already a huge gain.

I'm sorry to be blunt here, but 20ms is absolutely insane for non-crash events. This is pretty much guaranteed to cause stutters. This is a deal-breaker for most games out there, especially if we can't opt out of non-crash events

limbonaut:
@RafaelVidaurre The feature is generally opt-in. However, I believe that most users would prefer not to use it if we leave it as is. There is a bug in the latest release that affects even opted-out builds. We will release a patched version to address it soon.

bitsandfoxes:
In the event of an error, the SDK should capture and as detailed report as possible. With `CaptureScreenshot` as `opt-in` this means capturing a screenshot as well. I think the way forward is not to straight up drop screenshot capture in case of non-crash events but to provide more control. There I see two options:
1. A `BeforeCaptureScreenshot` callback so users can skip capturing based on their own logic. We could pass in the event.
2. More dedicated screenshot-options. I.e. some sort of eventlevel - screenshot capture matrix.

limbonaut:
> There I see two options

I believe option 1 might be a bit too much for something that could be represented as a flag. I’ve added an [issue](https://github.com/getsentry/sentry-godot/issues/149) tracking proposal similar to option 2. The screenshot-dropping part is removed from this PR, so we can move this along. What's left is performance improvements.

bitsandfoxes (approved):
Thanks!

Issue #150, opened 2025-03-27
Add `before_capture_screenshot` callback


limbonaut:
Implement `before_capture_screenshot` callback that should return a boolean value. If returned `false`, the screenshot capture should be omitted.

 _Originally posted by @bitsandfoxes in [#149](https://github.com/getsentry/sentry-godot/issues/149#issuecomment-2757602914)_

PR #151, opened 2025-03-27
feat: Allow enabling screenshots for specific event types

### LINKED ISSUE #149: Allow enabling `attach_screenshot` option separately for crash and non-crash events
Allow enabling `attach_screenshot` feature separately for crash and non-crash events. Currently, the `attach_screenshot` feature incurs performance overhead of saving a JPG image from the root viewport, which can take over 20 ms. This is way over a typical frame-time budget of 16 ms, and as such it introduces stuttering when an error occurs, as script and native errors can happen often in user projects.

Related discussion:
- https://github.com/getsentry/sentry-godot/pull/148

`before_capture_screenshot` is tracked in #150

limbonaut:
- Introduce event types in the public API:
  - Crash events
  - Error events *(non-crash events coming from the error logger)*
  - Custom events *(non-crash events created by users, such as those produced with the `capture_message()` method)*
- Allow screenshots to be enabled for specific event types.
- Event type can be checked in scripting too (e.g. in `before_send`).

Resolves #149

`before_capture_screenshot` will be covered in the follow-up: https://github.com/getsentry/sentry-godot/issues/150

#### In Project Settings

image

#### In code
```gdscript
options.screenshot_event_types = SentryEvent.EVENT_MASK_ALL
# or
options.screenshot_event_types = SentryEvent.EVENT_MASK_CRASH_EVENTS | SentryEvent.EVENT_MASK_CUSTOM_EVENTS
```

#### Tasks

- [x] Add tests
- [ ] Update online docs


src/sentry_options.cpp
```diff
@@ -64,6 +64,7 @@ void SentryOptions::_define_project_settings(const Ref<SentryOptions> &p_options
 
 	_define_setting("sentry/options/attach_log", p_options->attach_log);
 	_define_setting("sentry/options/attach_screenshot", p_options->attach_screenshot);
+	_define_setting(PropertyInfo(Variant::INT, "sentry/options/screenshot_event_mask", godot::PROPERTY_HINT_FLAGS, "Custom events,Crash events,Error events"), p_options->screenshot_event_mask);
```
limbonaut, line 67: I'm considering combining both options for screenshots into a single one.

src/sentry_options.cpp
```diff
@@ -64,6 +64,7 @@ void SentryOptions::_define_project_settings(const Ref<SentryOptions> &p_options
 
 	_define_setting("sentry/options/attach_log", p_options->attach_log);
 	_define_setting("sentry/options/attach_screenshot", p_options->attach_screenshot);
+	_define_setting(PropertyInfo(Variant::INT, "sentry/options/screenshot_event_types", godot::PROPERTY_HINT_FLAGS, "Custom events,Crash events,Error events"), p_options->screenshot_event_types);
```
limbonaut, line 67: I wonder if it makes sense to merge both screenshot options into a single one. Here’s an example:
```gdscript
options.attach_screenshot = SentryEvent.EVENT_MASK_ALL_EVENTS

# another example
options.attach_screenshot = SentryEvent.EVENT_MASK_CRASH_EVENTS | SentryEvent.EVENT_MASK_CUSTOM_EVENTS

```

src/sentry_options.cpp
```diff
@@ -64,6 +64,7 @@ void SentryOptions::_define_project_settings(const Ref<SentryOptions> &p_options
 
 	_define_setting("sentry/options/attach_log", p_options->attach_log);
 	_define_setting("sentry/options/attach_screenshot", p_options->attach_screenshot);
+	_define_setting(PropertyInfo(Variant::INT, "sentry/options/screenshot_event_types", PROPERTY_HINT_FLAGS, "Custom events,Crash events,Error events"), p_options->screenshot_event_types);
```
limbonaut, line 67: I wonder if it makes sense to merge both screenshot options into a single one. For example:
```gdscript
options.attach_screenshot = SentryEvent.EVENT_MASK_ALL_EVENTS

# another example
options.attach_screenshot = SentryEvent.EVENT_MASK_CRASH_EVENTS | SentryEvent.EVENT_MASK_CUSTOM_EVENTS

```

project/test/suites/test_options.gd
```diff
@@ -50,6 +51,14 @@ func test_max_breadcrumbs() -> void:
 	assert_int(options.max_breadcrumbs).is_equal(42)
 
 
+## SentryOptions.screenshot_event_types must be set to the specified value.
+func test_screenshot_event_types() -> void:
+	options.screenshot_event_types = SentryEvent.EVENT_MASK_CRASH_EVENTS | SentryEvent.EVENT_MASK_CUSTOM_EVENTS
+	assert_int(options.screenshot_event_types & SentryEvent.EVENT_MASK_CRASH_EVENTS).is_not_zero()
+	assert_int(options.screenshot_event_types & SentryEvent.EVENT_MASK_CUSTOM_EVENTS).is_not_zero()
```
bruno-garcia, line 58: What's custom events? We don't use this terminology in Sentry
How do you tell apart an error a crash and a 'custom'?

bruno-garcia (commented):
What is the default value?
Only crashes is what we use in other SDKs so I'd go with that for consistency sake

project/test/suites/test_options.gd
limbonaut: Events initiated by the user, like when you call SentrySDK.create_message() or `create_event()` + `capture_event()`. We can also call such events "manual" or "user" events if you like. Crash events are terminal events leading to app termination, initiated by crashpad in response to Godot app crash. The other two categories are non-terminating events. For example, errors in Godot do not lead to application termination. Godot tries to keep going as long as it can until an unrecoverable condition is met. Even division by zero is just a log message in the scripting engine turning numbers into NaNs.

> Only crashes is what we use in other SDKs so I'd go with that for consistency sake

Makes sense to me.

limbonaut:
> What is the default value?

Currently, all of those. I also lean to changing it to crash-only, as it is a more sensible default considering the performance implications.

bruno-garcia:
also, by default, there are no screenshots, right?
Since we don't know if the page has sensitive data (payment screen?) we can't automatically take screenshots.

Users must 'opt-in'. We usually have a simply `attachScreenshot` flag to make it easier. Options for types should be an 
additional option

limbonaut:
> also, by default, there are no screenshots, right?
> Users must 'opt-in'. We usually have a simply attachScreenshot flag to make it easier. Options for types should be an
additional option

We have `attach_screenshot` option that defaults to OFF. The whole problem right now, is that screenshots are expensive and introduce stutter when enabled for all events. Basically, we shouldn't default to taking them for non-crash events when `attach_screenshot` is enabled unless user specifically sets it so.

So the way I see it, it's either have some toggles with event categories, like errors/crashes/etc, or adding an option for a minimum event level, and if I understand it correctly, the fatal level is exactly the level for unrecoverable events.

PR #155, opened 2025-04-02
ref: Restructure options

limbonaut:
This PR restructures options in the Project Settings:
- Mark options as basic and advanced to have a cleaner interface.
- Move error logger tunables into their own sub-page. This is a **BREAKING** change so make sure to reapply those error logger values if you're changing the defaults.
  - Rename options: `error_logger_*` => `logger_*`.

#### Basic options

Screenshot 2025-04-04 at 17 40 51
Screenshot 2025-04-04 at 17 41 02

#### Advanced options

Screenshot 2025-04-04 at 17 41 27

Screenshot 2025-04-04 at 17 41 42

#### Options API

Screenshot 2025-04-04 at 17 48 02

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Breaking changes
 
 - Renamed `debug_verbosity` => `diagnostic_level` to better align with established Sentry features ([#154](https://github.com/getsentry/sentry-godot/pull/154))
+- Mark options as basic and advanced to have a cleaner interface, and move error logger tunables into their own sub-page. This is a BREAKING change so make sure to reapply those error logger values if you're changing the defaults. ([#155](https://github.com/getsentry/sentry-godot/pull/155))
 
 ### Features
 
--- a/doc_classes/SentryOptions.xml
+++ b/doc_classes/SentryOptions.xml
@@ -71,7 +71,7 @@
 		<member name="logger_event_mask" type="int" setter="set_logger_event_mask" getter="get_logger_event_mask" enum="SentryOptions.GodotErrorMask" is_bitfield="true" default="13">
 			Specifies the types of errors captured as events. Accepts a single value or a bitwise combination of [enum GodotErrorMask] masks.
 		</member>
-		<member name="logger_include_source" type="bool" setter="set_logger_include_source" getter="should_logger_include_source" default="true">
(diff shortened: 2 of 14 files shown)


bitsandfoxes (commented):
This looks so much cleaner now! 🔥 

I'm assuming a user here: Just from the option's name `"Error Logger" -> "Enabled"` it's not clear to me what that is or what it does. Does Sentry use one? Is it replacing Godot's? 
What about naming it `(Automatic?) Log Capture`, or `Logging Integration`?
Also, can we add a text-field to the options? Could be just a one-liner: "The SDK reads the log file (or hooks up into the logger in the future), parses log messages and sends those errors to Sentry".

limbonaut:
@bitsandfoxes Name is constructed from the setting keys we supply, but we can provide any key we want there, for example: 
- `sentry/error_logger/error_logger_enabled` key will transform to `"Error Logger" page -> "Error Logger Enabled" field`

We don't need to have the same names as in `SentryOptions` class, so there is some flexibility, but it's a good idea to keep it synchronized. We group all error logger related options under the `error_logger_` prefix in the options. This way, it's easy to communicate relation between them, or put them under the same category/group (we need to have a common prefix for this). I'm open to shortening it to `logger_*` if we want shorter names. `Log Capture` would clash with `Attach Log File` which would be confusing. And since we're expecting a `Logger` class [to be exposed in Godot](https://github.com/godotengine/godot/pull/91006/files#diff-0b36b3b54d28ebf01241affa02c98cd792676198da8a703c4b3cdd1ce5e784b4), I recommend sticking with the `Logger` prefixed names here. As we're keeping it compatible with the expected Godot changes, it will be a straightforward replacement when this functionality lands in Godot, reusing the same option keys.

As for some kind of a hint field, or text field, there is no such built-in option. I think we could develop something like that, but it's certainly wouldn't be trivial, as it would involve making an Inspector Editor plugin and property editor classes: maybe 100+ lines just to be able to do it in places. It may be possible, and I can look into it if you think it's worth exploring this idea.

In conclusion, I see the point of just `Enabled` being confusing, as it needs a little bit more context - I'll address that, keeping to the common prefix. And maybe shorten the error logger option keys to simply starting with `logger_`?

limbonaut:
I've updated the hat with the new screenshots. `logger_enabled` option has improved text and `error_logger_*` options shortened to `logger_*` . I also added a screenshot from the class docs showing how the API looks like with these changes.

bruno-garcia (approved):

PR #156, opened 2025-04-08
feat: Sign macOS builds

limbonaut:
Sign macOS binaries in CI. The signing process is only executed if the required secrets are accessible. On a release branch, the signing process is mandatory and will terminate with an error if the necessary keys are not provided via secrets.

EDIT: Code signing is now performed on release branch only.


limbonaut:
Tested and it works. Codesign is performed on release branch only.

bruno-garcia (approved):

PR #169, opened 2025-05-01
feat: Initial Android support

### LINKED ISSUE #89: Support Android targets
Sentry has extensive support for Android, including NDK support. 

This ticket tracks support for Android by relying on our Android native mobile SDK.

limbonaut:
This PR brings initial Android support. 

Since we need Callables in the Android Godot plugin API, this PR requires updating the minimum Godot version required to 4.4. This is a **breaking change**.

Resolves #89

Related issue:
- https://github.com/getsentry/sentry-java/issues/4479

**Known issues:**
- Currently missing screenshot and view hierarchy support, which need refactoring for Android.
- ~~`pre_context` & `post_context` are not included with captured events due to sentry-java issue linked above.~~ *(fixed in sentry-java 8.14.0)*
- Unit tests can't be run in the exported project on Android, so I've added basic tests to the demo project for now.

**Next steps:**
- Update CI to produce Android builds: #194
- Add screenshots and view hierarchy on Android: #202 
- Automate the `sentry-java` dependency update in CI: #200 
- Improve demo project UX on a mobile screen: #196 
- Investigate feasibility of testing on Android in CI.


kisg:
Is there a reason why you go through to the Java side for capturing events instead of relying on sentry-native, which is included in sentry-android: https://docs.sentry.io/platforms/android/configuration/using-ndk/?

limbonaut:
> Is there a reason why you go through to the Java side for capturing events instead of relying on sentry-native, which is included in sentry-android: https://docs.sentry.io/platforms/android/configuration/using-ndk/?

Replied [in the issue](https://github.com/getsentry/sentry-godot/issues/89#issuecomment-2893901629).

android_lib/build.gradle.kts
```diff
@@ -0,0 +1,63 @@
...
+        jvmTarget = "17"
+    }
+}
+
+dependencies {
+    implementation("org.godotengine:godot:4.4.0.stable")
+
+    // Note: Must also be updated in sentry_editor_export_plugin.cpp.
+    implementation("io.sentry:sentry-android:8.13.2")
+    implementation("com.jakewharton.threetenabp:threetenabp:1.4.9")
```
bruno-garcia: THis library has a note:

> Attention: Development on this library is winding down. Please consider switching to Android Gradle plugin 4.0, java.time.*, and its [core library desugaring](https://developer.android.com/studio/write/java8-support#library-desugaring) feature in the coming months.

Should we be looking into that instead? Or is there a way we can vendor some code and avoid the dependency altogether?

android_lib/build.gradle.kts
```diff
@@ -0,0 +1,63 @@
...
+        sourceCompatibility = JavaVersion.VERSION_17
+        targetCompatibility = JavaVersion.VERSION_17
+    }
+    kotlinOptions {
+        jvmTarget = "17"
+    }
+}
+
+dependencies {
+    implementation("org.godotengine:godot:4.4.0.stable")
```
bruno-garcia, line 37: To use a custom version of Godot (say 4.5) I'd need to change this? 
I imagine we'll need to adjust instructions about this.

android_lib/src/main/java/io/sentry/godotplugin/SentryAndroidGodotPlugin.kt
```diff
@@ -0,0 +1,376 @@
...
+class SentryAndroidGodotPlugin(godot: Godot) : GodotPlugin(godot) {
+    companion object {
+        private const val TAG = "sentry-godot"
+    }
+
+    private val eventsByHandle = ThreadLocal.withInitial { mutableMapOf<Int, SentryEvent>() }
+    private val exceptionsByHandle = ThreadLocal.withInitial { mutableMapOf<Int, SentryException>() }
+
+    private fun getEvent(eventHandle: Int): SentryEvent? {
+        val event: SentryEvent? = eventsByHandle.get()!![eventHandle]
```
bruno-garcia: ~If `eventsByHandle.get()` is null, it'll throw because of the `!!`, but we have a null check and a log below. So the indexer syntax can return null?
should we replace `!!` with `?.` instead?~
```suggestion
        val event: SentryEvent? = eventsByHandle.get()?.[eventHandle]
```

why. do we need `!!` to access `eventsByHandle` if we instantiate it with `withInitial`?

android_lib/src/main/java/io/sentry/godotplugin/SentryAndroidGodotPlugin.kt
```diff
@@ -0,0 +1,376 @@
...
+        if (event == null) {
+            Log.e(TAG, "Internal Error -- SentryEvent not found: $eventHandle")
+        }
+        return event
+    }
+
+    private fun getException(exceptionHandle: Int): SentryException? {
+        val exception: SentryException? = exceptionsByHandle.get()!![exceptionHandle]
```
bruno-garcia: same as above

android_lib/src/main/java/io/sentry/godotplugin/SentryAndroidGodotPlugin.kt
```diff
@@ -0,0 +1,376 @@
...
+        return handle
+    }
+
+    override fun getPluginName(): String {
+        return "SentryAndroidGodotPlugin"
+    }
+
+    override fun onGodotSetupCompleted() {
+        super.onGodotSetupCompleted()
+        AndroidThreeTen.init(this.activity!!.application) // needed for date time zones
```
bruno-garcia: can we vendor some small code instead of adding the dependency?

bruno-garcia: I wouldn't be surprised if we didn't have an activity instance. When we're responding to a push notification or some weird cases
Avoid the `!!` in favor of null checks instead

android_lib/src/main/java/io/sentry/godotplugin/SentryAndroidGodotPlugin.kt
```diff
@@ -0,0 +1,376 @@
...
+        Log.v(TAG, "Initializing Sentry Android")
+        SentryAndroid.init(godot.getActivity()!!.applicationContext) { options ->
+            options.dsn = dsn.ifEmpty { null }
+            options.isDebug = debug
+            options.release = release.ifEmpty { null }
+            options.dist = dist.ifEmpty { null }
+            options.environment = environment.ifEmpty { null }
+            options.sampleRate = sampleRate.toDouble()
+            options.maxBreadcrumbs = maxBreadcrumbs
+            options.sdkVersion!!.name = "sentry.java.android.godot"
```
bruno-garcia: Note we also need to add `sentry.native.android.godot` at the NDK layer
e.g:

https://github.com/getsentry/sentry-unreal/blob/e2a7b64f31ab0a3d8b43c7a36b686d9351a819a9/scripts/build-android.sh#L10

https://github.com/getsentry/sentry-unity/blob/c97e63f04fcb559f3740556041f8a3b497bbec3c/Directory.Build.targets#L239

android_lib/src/main/java/io/sentry/godotplugin/SentryAndroidGodotPlugin.kt
```diff
@@ -0,0 +1,376 @@
...
+            options.beforeSend =
+                SentryOptions.BeforeSendCallback { event: SentryEvent, hint: Hint ->
+                    Log.v(TAG, "beforeSend: ${event.eventId} isCrashed: ${event.isCrashed}")
+                    val handle: Int = registerEvent(event)
+                    Callable.call(beforeSendHandlerId, "before_send", handle)
+                    eventsByHandle.get()!!.remove(handle) // Returns the event or null if it was discarded.
```
bruno-garcia: how about a try/finally here to make sure we unregister/remove in case the callback throws an exception?

android_lib/src/main/java/io/sentry/godotplugin/UtilityFunctions.kt
```diff
@@ -0,0 +1,37 @@
...
+        SentryLevel.ERROR -> 3
+        SentryLevel.FATAL -> 4
+    }
+
+fun String.parseTimestamp(): Date {
+    val offsetDateTime = OffsetDateTime.parse(this)
+    return Date(offsetDateTime.toInstant().toEpochMilli())
+}
+
+fun Date.toRfc3339(): String {
```
bruno-garcia, line 101: Godot expects date in some specific format? I see we have 2 formats here and 1 of them means we need a library dependency

src/editor/sentry_editor_export_plugin.cpp
```diff
@@ -0,0 +1,30 @@
...
+PackedStringArray SentryEditorExportPlugin::_get_android_libraries(const Ref<EditorExportPlatform> &p_platform, bool p_debug) const {
+	PackedStringArray libs;
+	String release_or_debug = p_debug ? "debug" : "release";
+	libs.append("sentry/bin/android/sentry_android_godot_plugin." + release_or_debug + ".aar");
+	return libs;
+}
+
+PackedStringArray SentryEditorExportPlugin::_get_android_dependencies(const Ref<EditorExportPlatform> &p_platform, bool p_debug) const {
+	PackedStringArray deps;
+	// Note: Must be also updated in build.gradle.kts.
```
bruno-garcia:
```suggestion
	// NOTE: All dependencies below must be also updated in build.gradle.kts.
```

bruno-garcia (commented):
did 70% of the PR

android_lib/build.gradle.kts
limbonaut: I tried with desugaring first, but it requires changing user projects (manual steps needed to be performed by users), or the export process will fail to compile with obscure errors. My thinking is the dependency is a lesser evil here. We could try implementing our own version, if dependencies are not acceptable.

android_lib/build.gradle.kts
limbonaut: I don't think so, I think it's forward compatible within 4.x. We'll have to test this with 4.5.

UPDATE:
I've checked: I set the dependency to `org.godotengine:godot:4.3.0.stable` and the project ran fine with Godot 4.4.1 (except that before_send didn't work since it needs the newer dep).

android_lib/src/main/java/io/sentry/godotplugin/SentryAndroidGodotPlugin.kt
limbonaut: ThreadLocal.get() returns nullable. So we need either !! or ?.
 I used !! as I am fairly confident it exists.

limbonaut: Addressed in 3ea0077588534e7de3fd6dd242ec03350f9ca482

android_lib/src/main/java/io/sentry/godotplugin/SentryAndroidGodotPlugin.kt
limbonaut: Addressed in 3ea0077588534e7de3fd6dd242ec03350f9ca482

android_lib/src/main/java/io/sentry/godotplugin/SentryAndroidGodotPlugin.kt
limbonaut: Addressed in 3ea0077588534e7de3fd6dd242ec03350f9ca482

android_lib/src/main/java/io/sentry/godotplugin/SentryAndroidGodotPlugin.kt
limbonaut: No need, as It doesn't throw because it's a native call wrapper:
```java
@JvmStatic
fun call(godotObjectId: Long, methodName: String, vararg methodParameters: Any): Any? {
	return nativeCallObject(godotObjectId, methodName, methodParameters)
}

@JvmStatic
private external fun nativeCallObject(godotObjectId: Long, methodName: String, params: Array<out Any>): Any?

```

src/editor/sentry_editor_export_plugin.cpp
limbonaut: Addressed in ef586fcd8288b8d75e817fdf6dc18f2803c7c87c

android_lib/src/main/java/io/sentry/godotplugin/UtilityFunctions.kt
limbonaut: Godot has no native date or time type. This is all it has: https://docs.godotengine.org/en/latest/classes/class_time.html. So we currently use a formatted string as a timestamp, using RFC3339 format. Conveniently, that's what sentry-native returns for event.timestamp. I'm open to refactoring this to introduce a `SentryTimestamp` class.

limbonaut: Wdyt @bruno-garcia ?

android_lib/build.gradle.kts
limbonaut: Addressed in 06c9395fe974246fa3ddb9605a63dd02a2a1d0f9

android_lib/src/main/java/io/sentry/godotplugin/SentryAndroidGodotPlugin.kt
limbonaut: Addressed in 06c9395fe974246fa3ddb9605a63dd02a2a1d0f9

android_lib/src/main/java/io/sentry/godotplugin/SentryAndroidGodotPlugin.kt
limbonaut: Addressed in 2badfc3c82258273b33dc58c908e5e7b84d1df01

android_lib/src/main/java/io/sentry/godotplugin/SentryAndroidGodotPlugin.kt
```diff
@@ -0,0 +1,385 @@
...
+import kotlin.random.Random
+
+
+@Suppress("unused")
+class SentryAndroidGodotPlugin(godot: Godot) : GodotPlugin(godot) {
+    companion object {
+        private const val TAG = "sentry-godot"
+    }
+
+    private val eventsByHandle = ThreadLocal.withInitial { mutableMapOf<Int, SentryEvent>() }
```
romtsn: [withInitial](https://developer.android.com/reference/java/lang/ThreadLocal#withInitial(java.util.function.Supplier%3C?%20extends%20S%3E)) is only available from API 26, I noticed our minSdk is 24, so we should reconsider this probably

android_lib/src/main/java/io/sentry/godotplugin/SentryAndroidGodotPlugin.kt
```diff
@@ -0,0 +1,385 @@
...
+            Log.e(TAG, "Internal Error -- exceptionsByHandle is null")
+            return
+        }
+
+        exceptionsMap.remove(exceptionHandle)
+    }
+
+    @UsedByGodot
+    fun exceptionAppendStackFrame(exceptionHandle: Int, frameData: Dictionary) {
+        val exception = getException(exceptionHandle) ?: return
```
romtsn, line 366: so RN and Flutter switched over to building envelopes in the hybrid layer completely and then send raw bytes over the bridge to later call `InternalSentrySdk.captureEnvelope`. Is anything preventing us from doing the same thing here? Ideally we'd align between the hybrid SDKs wherever possible (cc @bruno-garcia).

Example from RN: https://github.com/getsentry/sentry-react-native/blob/c352c3096f425dbdf6864801b3aeba3f8b615712/packages/core/android/src/main/java/io/sentry/react/RNSentryModuleImpl.java#L548-L559

android_lib/src/main/java/io/sentry/godotplugin/SentryAndroidGodotPlugin.kt
limbonaut: Thanks, fixed in 2fd1eeaca52d207512917ab2297de39028388b34

android_lib/src/main/java/io/sentry/godotplugin/SentryAndroidGodotPlugin.kt
limbonaut: I have a few concerns. `sentry-native` doesn't seem to support constructing and capturing envelopes, so this functionality would be a separate code path just for Android support (cc @vaind @jpnurmi).
The other one, `InternalSentrySdk::captureEnvelope()` [claims](https://github.com/getsentry/sentry-java/blob/41527016cbee901e779c1383170a185b14e71e1b/sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java#L152) that events are not enriched with additional data. We would prefer to have Android contexts/scopes applied to such error events, right? For the context, these methods are mostly used to report script errors happening at runtime such as GDScript errors.

romtsn: > I have a few concerns. sentry-native doesn't seem to support constructing and capturing envelopes, so this functionality would be a separate code path just for Android support (cc @vaind @jpnurmi).

Do you mean as public API? Because it works that way on Android - captures envelopes and stores them to the `cache/outbox` folder so they get picked up and sent by the Android SDK.

>  The other one, InternalSentrySdk::captureEnvelope() [claims](https://github.com/getsentry/sentry-java/blob/41527016cbee901e779c1383170a185b14e71e1b/sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java#L152) that events are not enriched with additional data. We would prefer to have Android contexts/scopes applied to such error events, right?

Right, that's because we have a separate [serializeScope](https://github.com/getsentry/sentry-java/blob/41527016cbee901e779c1383170a185b14e71e1b/sentry-android-core/src/main/java/io/sentry/android/core/InternalSentrySdk.java#L79-L147) method now which also should be called manually from the hybrid SDKs.

I don't have a strong opinion here, just thought it makes sense to at least start the conversation.

limbonaut: > Do you mean as public API?

Yeah I mean we capture script errors on desktop platforms, and we'll do the same on Android.

> I don't have a strong opinion here, just thought it makes sense to at least start the conversation.

If passing events as an envelop means we'll get better performance, it's probably worth looking into.

android_lib/build.gradle.kts
```diff
@@ -0,0 +1,63 @@
...
+    alias(libs.plugins.android.library)
+    alias(libs.plugins.kotlin.android)
+}
+
+android {
+    namespace = "io.sentry.godotplugin"
+    compileSdk = 35
+
+    defaultConfig {
+        minSdk = 24
```
bruno-garcia: is this the lowest version Godot supports?
[Our SDK supports 21+](https://github.com/getsentry/sentry-java#sentry-sdk-for-java-and-android)

android_lib/build.gradle.kts
```diff
@@ -0,0 +1,63 @@
...
+        jvmTarget = "17"
+    }
+}
+
+dependencies {
+    implementation("org.godotengine:godot:4.4.0.stable")
+    testImplementation("junit:junit:4.13.2")
+
+    // NOTE: All dependencies below must be also updated in sentry_editor_export_plugin.cpp.
+    implementation("io.sentry:sentry-android:8.14.0")
```
bruno-garcia, line 41: we'll need an update dep automation for this. See: https://github.com/getsentry/github-workflows

android_lib/src/main/java/io/sentry/godotplugin/SentryAndroidGodotPlugin.kt
```diff
@@ -0,0 +1,394 @@
...
+    }
+
+    @UsedByGodot
+    fun setContext(key: String, value: Dictionary) {
+        Sentry.getGlobalScope().setContexts(key, value)
+    }
+
+    @UsedByGodot
+    fun removeContext(key: String) {
+        // Q: Godot SDK doesn't expose scopes yet, should we set it on the global scope?
```
bruno-garcia: yes, just global since Godot is a desktop/mobile app and we default to global mode in these scenarios.

thread-local is the default on server apps

android_lib/src/main/java/io/sentry/godotplugin/SentryAndroidGodotPlugin.kt
```diff
@@ -0,0 +1,394 @@
...
+    }
+
+    @UsedByGodot
+    fun eventSetRelease(eventHandle: Int, release: String) {
+        getEvent(eventHandle)?.release = release
+    }
+
+    @UsedByGodot
+    fun eventGetRelease(eventHandle: Int): String {
+        return getEvent(eventHandle)?.release ?: ""
```
bruno-garcia, line 300: can we return `null` instead? Not sure about returning a value `""` all over, it'll show in Sentry as having the key and the vallue `""` where in fact we should avoid setting the key to begin with, if there's no value

android_lib/src/main/java/io/sentry/godotplugin/SentryAndroidGodotPlugin.kt
bruno-garcia: the public API can be one thing, while the internals can be another.
There's value in aligning with other hybrid SDKs, but why did we take that route @romtsn , was it due to possible incompatibilities?

The event here is quite simple so this approach for Godot seems OK for now.
But we should look at what the other hybrid ones are doing and consider if aligning makes sense.


That said, we also have Unity, Unreal and .NET binding to Android so we could look at those too (iirc we don't do envelope and bytes on any of those)

bruno-garcia (approved):
lgtm, we can follow up on open comments in other PRs because this is quite big already

android_lib/build.gradle.kts
limbonaut: Right, it should be 21. In Godot 4.5, IIRC it's going up to 24.
Addressed in 201ebcea153426aa00a8e601ca8bcd90ba20114b.

android_lib/build.gradle.kts
limbonaut: I started on it in #200. I'll take a look at the workflows repo.

android_lib/src/main/java/io/sentry/godotplugin/SentryAndroidGodotPlugin.kt
limbonaut: Godot doesn't support nullable types, so we can't pass back null. However, I don't think it's a problem, because the setter is only called when the user specifically calls setRelease, so it's going to stay null unless specifically set to an empty string by the caller.

android_lib/src/main/java/io/sentry/godotplugin/SentryAndroidGodotPlugin.kt
romtsn: it's all in here: https://github.com/getsentry/team-mobile/issues/11, but the main reason is - to have device context (and whatever comes from the native SDKs) available in the Hybrid's beforeSend callback.

PR #186, opened 2025-06-03
ref: Improve thread safety

limbonaut:
* Store last_uuid as atomic
* Synchronize access to SentryUser
* Add test for threaded event capture
* Add `SentryUser.duplicate()`

### DIFF
--- a/src/sentry_sdk.cpp
+++ b/src/sentry_sdk.cpp
@@ -101,7 +101,7 @@ void SentrySDK::set_user(const Ref<SentryUser> &p_user) {
 
 Ref<SentryUser> SentrySDK::get_user() const {
 	MutexLock lock(*user_mutex.ptr());
-	return user;
+	return user->duplicate();
 }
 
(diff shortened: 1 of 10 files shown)


tustanivsky (approved):

src/sentry_sdk.cpp
```diff
@@ -96,7 +99,13 @@ void SentrySDK::set_user(const Ref<SentryUser> &p_user) {
 	}
 }
 
+Ref<SentryUser> SentrySDK::get_user() const {
+	MutexLock lock(*user_mutex.ptr());
```
bruno-garcia, line 103: I assume ppl would use this method to mutate the user. now they will have a clone. not sure this is the way to go

limbonaut: They have a clone of the user data. They are supposed to use `set_user()` to finalize their changes. Or should we support editing user properties in-place and picking up those changes automatically without a call to `set_user()`?

PR #187, opened 2025-06-03
ref: Implement Event.add_exception() to replace capture_error()

limbonaut:
This is needed for both godot-4.5 branch and android.

#skip-changelog since it's a pure refactor


tustanivsky (approved):
LGTM!

PR #189, opened 2025-06-05
fix: Filter out expected missing attachment warnings

limbonaut:
Remove messages from logging related to missing `screenshot.jpg` and `view-hierarchy.json` attachments. They are expected to be missing in some scenarios.


tustanivsky (approved):
Nice! We may need to add something similar in Unreal.

PR #191, opened 2025-06-06
Fix crash in code editor when `disabled_in_editor` is ON

limbonaut:
Fix the crash that can happen if you try to write such code in GDScript code editor:
```gdscript
var ev := SentrySDK.create_event()
ev.message = "some message"
```


tustanivsky (approved):

limbonaut:
Thank you for your prompt review, @tustanivsky 😊

PR #194, opened 2025-06-11
ci: Make Android builds

limbonaut:
Update CI to build and package Android targets.

* Depends on #169

#skip-changelog

### DIFF
--- a/.github/workflows/build_gdextension.yml
+++ b/.github/workflows/build_gdextension.yml
@@ -150,6 +150,13 @@ jobs:
           distribution: temurin
           java-version: 17
 
+      - name: Set up NDK
+        if: matrix.platform == 'android'
+        uses: nttld/setup-ndk@afb4c9964b521afb97c864b7d40b11e6911bd410 # v1.5.0
+        with:
+          ndk-version: r27c
+          link-to-sdk: true
+
       - name: Compile GDExtension library
         shell: bash
         run: |
--- a/.github/workflows/build_gdextension.yml
+++ b/.github/workflows/build_gdextension.yml
@@ -178,7 +178,7 @@ jobs:
         if: matrix.platform != 'web' # web is unsupported currently; it is used to test compilation with such platforms (ifdefs)
         uses: actions/upload-artifact@v4
         with:
-          name: libsentry.${{matrix.platform}}.${{matrix.target}}.${{matrix.arch}}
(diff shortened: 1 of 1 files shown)


.github/workflows/build_gdextension.yml
```diff
@@ -128,20 +178,48 @@ jobs:
...
+
+      - name: Upload artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: sentry.android.godot_plugin
+          path: |
+            project/
+            !project/
+            project/addons/sentry/bin/android/sentry_android_godot_plugin.debug.aar
+            project/addons/sentry/bin/android/sentry_android_godot_plugin.release.aar
```
jpnurmi, line 213: is there some advantage with this syntax? is the result not the same as with only listing the two .aar files?

.github/workflows/build_gdextension.yml
```diff
@@ -107,6 +143,20 @@ jobs:
...
+      - name: Set up Java 17
+        if: matrix.platform == 'android'
+        uses: actions/setup-java@c5195efecf7bdfc987ee8bae7a71cb8b11521c00 # v4.7.1
+        with:
+          distribution: temurin
+          java-version: 17
+
+      - name: Set up NDK
+        if: matrix.platform == 'android'
+        uses: nttld/setup-ndk@afb4c9964b521afb97c864b7d40b11e6911bd410 # v1.5.0
```
jpnurmi, line 155: not sure what is "nttld" but did you check if there's anything more "official" available?

jpnurmi (commented):
Nice and tidy! :+1:

.github/workflows/build_gdextension.yml
limbonaut: Yeah, it's a trick to force include files starting with the `addons` folder, as opposed to just having those files at the root of the archive. Since we use merge action later, this way we don't have to manually move those files to proper location.

UPDATE:
Quoting the docs:
> If multiple paths are provided as input, the least common ancestor of all the search paths will be used as the root directory of the artifact. Exclude paths do not affect the directory structure.

.github/workflows/build_gdextension.yml
limbonaut: Not that I know of. Just checked again – nothing more official exists, looks like it.

PR #201, opened 2025-06-17
feat: Support for logger, script stack traces, local variables in Godot 4.5

limbonaut:
Add support for:
- **Logger interface** introduced in Godot 4.5, which enables more accurate error reporting.
- **Script stack traces** for GDScript and potentially other scripting languages
- **Local variables** support (opt-in feature, as it incurs a significant performance overhead)

This PR replaces the previous logger implementation with a log-file-based approach. As such, the minimum Godot version required will be 4.5.

Note for reviewers: sentry_logger.{h,cpp} are rewritten from scratch, so it's best to look at those as new files, not diff. 

script-stack-trace-editor


src/sentry_logger.cpp
```diff
@@ -4,300 +4,275 @@
 #include "sentry_options.h"
 #include "sentry_sdk.h"
```
limbonaut, line 5: Suggestion for reviewers: Best look at this file as a whole, not diff ("three dots" -> View file). It's almost new – it replaces the old implementation.

src/sentry_options.cpp
```diff
@@ -71,10 +71,10 @@ void SentryOptions::_define_project_settings(const Ref<SentryOptions> &p_options
 
 	_define_setting("sentry/logger/logger_enabled", p_options->logger_enabled);
 	_define_setting("sentry/logger/include_source", p_options->logger_include_source, false);
+	_define_setting("sentry/logger/include_variables", p_options->logger_include_variables, false);
 	_define_setting(PropertyInfo(Variant::INT, "sentry/logger/events", PROPERTY_HINT_FLAGS, sentry::GODOT_ERROR_MASK_EXPORT_STRING()), p_options->logger_event_mask, false);
 	_define_setting(PropertyInfo(Variant::INT, "sentry/logger/breadcrumbs", PROPERTY_HINT_FLAGS, sentry::GODOT_ERROR_MASK_EXPORT_STRING()), p_options->logger_breadcrumb_mask, false);
 
-	_define_setting("sentry/logger/limits/parse_lines", p_options->logger_limits->parse_lines, false);
```
limbonaut, line 77: Explanation: This option is not needed because we don't parse the log file anymore.

src/sentry_sdk.cpp
```diff
@@ -237,6 +268,11 @@ void SentrySDK::_check_if_configuration_succeeded() {
 	}
 }
 
+void SentrySDK::_demo_helper_crash_app() {
+	char *ptr = (char *)1;
+	sentry::util::print_fatal("Crash by access violation ", ptr); // this is going to crash the app
+}
+
```
limbonaut, line 275: Explanation: Need a new crasher method, as the old one generates an additional error in Godot 4.5.

bruno-garcia:
@sentry review

bruno-garcia:
bugbot run

src/sentry_logger.cpp
```diff
@@ -4,300 +4,275 @@
...
+		return true;
+	}
+
+	String source_code = script->get_source_code();
+	if (source_code.is_empty()) {
+		sentry::util::print_debug("Script source not available ", p_file.utf8().ptr());
+		return true;
 	}
 
+	PackedStringArray lines = script->get_source_code().split("\n");
```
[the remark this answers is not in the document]
limbonaut: The function has a clear, single purpose -- extracting script context from a file and line number. It doesn't need to be part of the class interface since it doesn't access any class members. Since it's only currently used withing this translation unit, the anonymous namespace is appropriate scope.

src/sentry_logger.cpp
```diff
@@ -4,300 +4,275 @@
...
+					int32_t num_vars = num_locals + num_members + num_globals;
+					int32_t starting_index = 0;
+
+					stack_frame.vars.resize(num_vars);
+
+					for (int i = 0; i < num_locals; i++) {
+						stack_frame.vars.set(starting_index + i,
+								Pair(backtrace->get_local_variable_name(frame_idx, i), backtrace->get_local_variable_value(frame_idx, i)));
+					}
+					starting_index += num_locals;
```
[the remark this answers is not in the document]
limbonaut: Hello, seer. There is no separate reserve() operation in the godot::Vector – `resize()` is the only reserve-like operation. This code is fine.

src/sentry_logger.cpp
```diff
@@ -4,300 +4,275 @@
...
+			within_throttling_limit &&
+			!is_spammy_error;
+	bool as_breadcrumb = SentryOptions::get_singleton()->should_capture_breadcrumb((GodotErrorType)p_error_type) &&
+			!is_spammy_error;
+
+	if (as_event) {
+		// We decided to capture the error as event (it's happening).
+		frame_events++;
+		event_times.push_back(now);
 	}
```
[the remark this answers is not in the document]
limbonaut: Don't mind using guards, but this reminds me of something. We use standard library mutexes in other places, and we should stick to the same methods here.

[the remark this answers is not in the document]
limbonaut: Addressed in 5b64e29b6edf826b8c1e2512606e5dbc49c8337a

src/sentry_logger.cpp
```diff
@@ -4,300 +4,275 @@
...
-		SentryEvent::StackFrame stack_frame{
-			p_file,
-			p_func,
-			p_line,
-			true, // in_app
-			"godot" // platform
-		};
+		Vector<SentryEvent::StackFrame> frames;
+
+		// Backtraces don't include variables by default, so if we need them, we must capture them separately.
```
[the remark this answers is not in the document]
limbonaut: I like the idea of checking file/line first. I don't think that prioritizing GDScript is particularly a good idea.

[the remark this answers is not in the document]
limbonaut: Addressed in ad21d57fe614f4325e4610ea72e899b4caab3e5a

doc_classes/SentryOptions.xml
```diff
@@ -77,6 +77,9 @@
 		<member name="logger_include_source" type="bool" setter="set_logger_include_source" getter="is_logger_include_source_enabled" default="true">
 			If [code]true[/code], the SDK will include the surrounding source code of logged errors, if available in the exported project.
 		</member>
+		<member name="logger_include_variables" type="int" setter="set_logger_include_variables" getter="is_logger_include_variables_enabled" default="0">
```
jpnurmi: should it be `type="bool"` and `default="false"`?

limbonaut: Thanks! Fixed in 0eaf2ecaf09821dce94507ba2ea91510e1080aa7.

jpnurmi:
`sentry_logger.cpp` looks good to me but is quite heavy. 😅 Does anyone else want to give it a pair of eyes?

.github/actions/prepare-testing/action.yml
```diff
@@ -108,5 +108,5 @@ runs:
       run: |
         find project/addons/sentry/bin/ -name "crashpad_handler" -exec chmod 755 "{}" \;
         echo "--- Rebuilding import cache"
-        ${GODOT} --headless --editor --path project/ --quit-after 2000 || true
+        ${GODOT} --headless --editor --path project/ --quit-after 2000
```
limbonaut, line 111: Godot used to crash here due to engine bug. This is no longer necessary.

src/sentry_logger.cpp
```diff
@@ -4,300 +4,287 @@
...
+					true, // in_app
+					platform
+				};
+
+				// Provide script source code context for script errors if available.
+				if (SentryOptions::get_singleton()->is_logger_include_source_enabled()) {
+					String context_line;
+					PackedStringArray pre_context;
+					PackedStringArray post_context;
+					bool err = _get_script_context(backtrace->get_frame_file(frame_idx),
```
tustanivsky: The return value in `_get_script_context` seems a bit counterintuitive - returning `true` on error feels opposite to what I’d expect

src/sentry_logger.cpp
```diff
@@ -4,300 +4,287 @@
...
 	// Capture error as event.
 	if (as_event) {
-		SentryEvent::StackFrame stack_frame{
-			p_file,
-			p_func,
-			p_line,
-			true, // in_app
-			"godot" // platform
-		};
+		Vector<SentryEvent::StackFrame> frames;
```
tustanivsky: Maybe we can extract the stack walking logic from here since this method is already quite large?

jpnurmi (approved):

src/sentry_logger.cpp
limbonaut: Addressed in b4ad75c0d67293e7487887b71f2ee76b287b4651

src/sentry_logger.cpp
limbonaut: Addressed in e5d064148a1208a32ca194c4f6db44a57d5e1b2c

limbonaut:
Thank you for reviewing @jpnurmi @tustanivsky

PR #202, opened 2025-06-17
ref: Event processing

limbonaut:
* Revise event processing to support Android. This fixes missing screenshots and view hierarchy on Android.
* Introduce event processors (internally). This interface can also be later exposed in the public API if needed.

Depends on:
- #169

#skip-changelog

### DIFF
--- a/src/sentry/processing/screenshot_processor.cpp
+++ b/src/sentry/processing/screenshot_processor.cpp
@@ -17,31 +17,26 @@ Ref<SentryEvent> ScreenshotProcessor::process_event(const Ref<SentryEvent> &p_ev
 		return nullptr;
 	}
 
-	bool is_main_thread = OS::get_singleton()->get_thread_caller_id() == OS::get_singleton()->get_main_thread_id();
+	if (!DisplayServer::get_singleton() || DisplayServer::get_singleton()->get_name() == "headless") {
+		return p_event;
+	}
+
+	int32_t current_frame = Engine::get_singleton()->get_frames_drawn();
 
 	{
 		std::lock_guard lock{ mutex };
 
-		int32_t current_frame = Engine::get_singleton()->get_frames_drawn();
 		if (current_frame == last_screenshot_frame) {
-			sentry::util::print_debug("skipping screenshot");
-			// Screenshot already exists for this frame — nothing to do.
+			sentry::util::print_debug("skipping screenshot – already taken");
 			return p_event;
 		}
-		if (is_main_thread) {
-			last_screenshot_frame = current_frame;
-		}
+
+		// Remove the outdated screenshot.
+		DirAccess::remove_absolute(screenshot_path);
 	}
 
-	String screenshot_path = "user://" SENTRY_SCREENSHOT_FN;
-	DirAccess::remove_absolute(screenshot_path);
-
-	if (!is_main_thread) {
-		sentry::util::print_debug("skipping screenshot capture - can only be performed on the main thread");
-		return p_event;
-	}
-
-	if (!DisplayServer::get_singleton() || DisplayServer::get_singleton()->get_name() == "headless") {
+	if (OS::get_singleton()->get_thread_caller_id() != OS::get_singleton()->get_main_thread_id()) {
+		sentry::util::print_debug("skipping screenshot – can only be performed on the main thread");
 		return p_event;
 	}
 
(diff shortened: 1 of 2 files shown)


limbonaut:
I'm not sure if we need a changelog entry for this.

bruno-garcia:
@sentry review

limbonaut:
I addressed points that I found useful from the bot's review.

bruno-garcia:
@romtsn maybe can peek at the kotlin? hoping @tustanivsky or @vaind can peek at the C++ 🙏

src/sentry/processing/screenshot_processor.cpp
```diff
@@ -0,0 +1,79 @@
...
+		std::lock_guard lock{ mutex };
+
+		int32_t current_frame = Engine::get_singleton()->get_frames_drawn();
+		if (current_frame == last_screenshot_frame) {
+			sentry::util::print_debug("skipping screenshot");
+			// Screenshot already exists for this frame — nothing to do.
+			return p_event;
+		}
+		if (is_main_thread) {
+			last_screenshot_frame = current_frame;
```
romtsn: I'm thinking this is done too early - e.g. if we exit early in one of the `ifs` below, like get_level or before_capture checks, we'd still memoize the `last_screenshot_frame` here although the screenshot hasn't been taken.

src/sentry/processing/view_hierarchy_processor.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+Ref<SentryEvent> ViewHierarchyProcessor::process_event(const Ref<SentryEvent> &p_event) {
+#ifdef DEBUG_ENABLED
+	uint64_t start = Time::get_singleton()->get_ticks_usec();
+#endif
+
+	String path = "user://" SENTRY_VIEW_HIERARCHY_FN;
+	DirAccess::remove_absolute(path);
+
+	if (OS::get_singleton()->get_thread_caller_id() != OS::get_singleton()->get_main_thread_id()) {
```
romtsn, line 20: wondering if there's a way in godot to actually post a message to the main thread where you could grab the view hierarchy instead of bailing out? That's what we do on Android: https://github.com/getsentry/sentry-java/blob/736a33bfcbdca13894c84818136fc9887ebddf4a/sentry-android-core/src/main/java/io/sentry/android/core/ViewHierarchyEventProcessor.java#L180-L188

romtsn (approved):
kotlin LGTM, a couple of questions to the cpp code, but I'm no expert there so take it or leave it :)

src/sentry/processing/view_hierarchy_processor.cpp
limbonaut: There is a message queue that we can use and initially that was my intention. However, the main issue with it is that until our request is processed, the scene tree is likely changed already. It's a volatile structure, nodes getting added and removed during frame processing. So we would be reporting potentially incorrect state and presenting it as the state in the moment of error. In such a case, I think it's best not to report it.

romtsn: > However, the main issue with it is that until our request is processed, the scene tree is likely changed already. It's a volatile structure, nodes getting added and removed during frame processing

is that done on a separate thread/queue? Because if it's being done on the main/UI thread that shouldn't be an issue, since you're going to be blocking it while capturing the view hierarchy snapshot

src/sentry/processing/screenshot_processor.cpp
limbonaut: Addressed in 61e0e58fd43c14e4be96e023be9581f985879c7b

src/sentry/processing/view_hierarchy_processor.cpp
limbonaut: Given that a view hierarchy can only be captured on the main thread, and we queue a message from a different thread, we will block the calling thread until the main thread processes the main loop and the message queue is flushed, and our request gets finally processed – the scene tree state will most likely not be the same. The problem I see is that the state we would represent is quite likely misleading than helpful.

(The message queue on the main thread is flushed at specific moments like [here](https://github.com/godotengine/godot/blob/e1b4101e3460dd9c6ba0b7f8d88e9751b8383f5b/main/main.cpp#L4739-L4774) and [here](https://github.com/godotengine/godot/blob/e1b4101e3460dd9c6ba0b7f8d88e9751b8383f5b/scene/main/scene_tree.cpp#L693-L700).)

romtsn: makes sense 👍

bruno-garcia (approved):

tustanivsky (commented):
C++ looks good as well - really like this new approach with event processors!

PR #204, opened 2025-06-19
feat: User feedback API

limbonaut:


src/sentry/native/native_sdk.cpp
```diff
@@ -340,6 +340,26 @@ String NativeSDK::capture_event(const Ref<SentryEvent> &p_event) {
 	return _uuid_as_string(uuid);
 }
 
+void NativeSDK::capture_feedback(const Ref<SentryFeedback> &p_feedback) {
+	ERR_FAIL_COND(p_feedback.is_null());
+	ERR_FAIL_COND(p_feedback->get_message().is_empty());
+
+	sentry_uuid_t event_uuid;
+	if (p_feedback->get_associated_event_id().is_empty()) {
+		event_uuid = sentry_uuid_nil();
```
aliu39, line 349: is this a random uuid? If the id doesn't correspond to an error event the feedback won't be created in sentry, if you're using the old `capture_user_feedback`

src/sentry/native/native_sdk.cpp
```diff
@@ -340,6 +340,26 @@ String NativeSDK::capture_event(const Ref<SentryEvent> &p_event) {
...
+		event_uuid = sentry_uuid_from_string(p_feedback->get_associated_event_id().ascii());
+	}
+
+	sentry_value_t uf = sentry_value_new_user_feedback(
+			&event_uuid,
+			p_feedback->get_name().utf8(),
+			p_feedback->get_contact_email().utf8(),
+			p_feedback->get_message().utf8());
+
+	sentry_capture_user_feedback(uf);
```
aliu39, line 360: why are we wrapping the old capture_user_feedback here? This will still send the old envelope format

bruno-garcia: The Godot SDK is built on top of sentry-native, which doesn't yet have the new `capture_feedback`.

See: https://github.com/getsentry/sentry-native/issues/885#issuecomment-2428023882

src/sentry/native/native_sdk.cpp
limbonaut: Thanks for bringing this up, I didn't realize it's a different protocol.

limbonaut:
Indeed, if it just gets attached to an event, it's not as useful for what we have in mind. It needs more cooking then.

limbonaut:
Superseded by #418

PR #205, opened 2025-06-19
feat: Add user attachments support

### LINKED ISSUE #120: Support arbitrary attachments
We currently support attaching only log files via `options.attach_log`

limbonaut:
Add user attachments support

- **Attachments**: Users can now add attachments with path or bytes on the global scope
- **Two creation methods**:
  - `SentryAttachment.create_with_path()` - attach files from disk 
  - `SentryAttachment.create_with_bytes()` - attach in-memory data
- Works on desktop and Android
- Resolves #120 
- Wishlist: https://github.com/getsentry/sentry-native/issues/1289
- Documentation PR: https://github.com/getsentry/sentry-docs/pull/14210


src/sentry_sdk.cpp
```diff
@@ -77,6 +78,18 @@ String SentrySDK::capture_event(const Ref<SentryEvent> &p_event) {
...
+void SentrySDK::add_attachment(const Ref<SentryAttachment> &p_attachment) {
+	ERR_FAIL_COND_MSG(p_attachment.is_null(), "Sentry: Can't add null attachment.");
+	ERR_FAIL_COND_MSG(p_attachment->get_path().is_empty(), "Sentry: Can't add attachment with empty path.");
+
+	internal_sdk->add_attachment(p_attachment);
+}
+
+void SentrySDK::remove_attachment(const Ref<SentryAttachment> &p_attachment) {
+	ERR_FAIL_COND_MSG(p_attachment.is_null(), "Sentry: Can't remove null attachment.");
+	internal_sdk->remove_attachment(p_attachment);
```
limbonaut: I noticed that sentry-cocoa doesn’t have the `scope.removeAttachment` method. Is this going to be an issue later on?

limbonaut:
@sentry review

android_lib/src/main/java/io/sentry/godotplugin/SentryAndroidGodotPlugin.kt
```diff
@@ -123,11 +124,44 @@ class SentryAndroidGodotPlugin(godot: Godot) : GodotPlugin(godot) {
...
+            false
+        )
         Sentry.getGlobalScope().addAttachment(attachment)
     }
 
+    @UsedByGodot
+    fun removeFileAttachment(path: String) {
+        val globalScope = Sentry.getGlobalScope()
+        val attachment = globalScope.attachments.find { att -> att.pathname == path }
+        globalScope.attachments.remove(attachment)
```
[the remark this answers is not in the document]
limbonaut: not using `attachment` as it coincides with the upper scope variable which would lead to confusion.

android_lib/src/main/java/io/sentry/godotplugin/SentryAndroidGodotPlugin.kt
```diff
@@ -123,11 +124,44 @@ class SentryAndroidGodotPlugin(godot: Godot) : GodotPlugin(godot) {
...
+        )
+        Sentry.getGlobalScope().addAttachment(attachment)
+    }
+
+    @UsedByGodot
+    fun removeBytesAttachment(filename: String) {
+        val globalScope = Sentry.getGlobalScope()
+        val attachment = globalScope.attachments.find { att -> att.filename == filename }
+        globalScope.attachments.remove(attachment)
+    }
```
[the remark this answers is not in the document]
limbonaut: unused variable fixed

src/sentry_attachment.h
```diff
@@ -0,0 +1,48 @@
...
+	String get_path() const { return path; }
+	String get_filename() const { return filename; }
+	String get_content_type() const { return content_type; }
+
+#ifdef NATIVE_SDK
+	sentry_attachment_t *get_native_attachment() const { return native_attachment; }
+	void set_native_attachment(sentry_attachment_t *p_native_attachment) { native_attachment = p_native_attachment; }
+#endif
+
+	virtual ~SentryAttachment() = default;
```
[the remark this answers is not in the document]
limbonaut: SentryAttachment is no longer a base class, so this is correct.

doc_classes/SentrySDK.xml
```diff
@@ -64,6 +72,13 @@
 				Returns whether the SDK is enabled, i.e. whether it is initialized and active.
 			</description>
 		</method>
+		<method name="remove_attachment">
+			<return type="void" />
+			<param index="0" name="attachment" type="SentryAttachment" />
+			<description>
+				Removes a previously added attachment from future Sentry events. The [param attachment] should be a [SentryAttachment] object returned by [method SentrySDK.add_attachment].
```
jpnurmi: should it say something like "added with" instead of "returned by"?

src/sentry_attachment.h
```diff
@@ -0,0 +1,48 @@
...
+
+#ifdef NATIVE_SDK
+	sentry_attachment_t *native_attachment = nullptr;
+#endif
+
+protected:
+	static void _bind_methods();
+
+public:
+	static Ref<SentryAttachment> create_with_path(const String &p_path, const String &p_filename = "", const String &p_content_type = "");
```
jpnurmi: Is there a reason such as immutability to avoid setters? This looks convenient at first sight but three string arguments in a row creates a risk of arguments getting mixed up and reduces readability. :)

```cpp
auto attachment = SentryAttachment.create_with_path("/path/to/file.bin", "", "application/octet-stream");
```
vs.
```cpp
auto attachment = SentryAttachment.create_with_path("/path/to/file.bin");
attachment.set_content_type("application/octet-stream");
```

doc_classes/SentrySDK.xml
limbonaut: Good catch – initially, it was returning the attachment object. Thanks!

limbonaut: Fixed in 58da2f6

src/sentry_attachment.h
limbonaut: The reason was that mutating these properties after `add_attachment` is called would have no actual effect. But maybe you're right with the "three strings" point.

limbonaut: Addressed in eea3098

src/sentry_sdk.cpp
tustanivsky: In cocoa, scope has a public `attachmentArray` property which you can modify directly to remove the attachment.

tustanivsky: UPD: I've just got to implementing this feature in Unreal SDK and it turns out that we can't actually access attachments as mentioned above (the header I was referencing earlier was from an internal `sentry-cocoa` module). Currently, the only option is to remove all attachments at once with `clearAttachments`.

Somewhat similar situation we have on Android:
```java
// Here `getAttachments` returns instance of `CopyOnWriteArrayList`
// which can't be used to modify the original
Sentry.getGlobalScope().getAttachments().remove(attachment);
```

limbonaut: Dang, this means it's not implemented on Android either. I wonder if removing attachment is actually a supported use-case? Maybe they are supposed to last for the whole scope's lifetime?
@bruno-garcia wdyt?

bruno-garcia: if I'm using the SDK with sentry.configureScope(s => s.addAttachment) on mobile I'd expect to stay for the duration of the app's lifecycle. If I want 1 attachment for 1 capture that I do manually, I'd do: captureEvent(event, s => s.addAttachment), or we use hints perhaps, I'm not sure what's the API but it's basically passing directly to the `captureXyz` method.

It's probably fine to add support to add without the ability to delete just yet if the native SDKs don't have that ability either. At least ClearAttachments should be possible though

limbonaut: I removed `remove_attachment()` for now, just so we can unblock this. It seems that `clear_attachments()` has some drawbacks, as it's unclear how it should handle something like view hierarchy file or screenshots. If it were to  remove such attachments, it's usefulness might be questionable.

PR #215, opened 2025-06-27
ci: Fix isolated tests routine executed for .uid files

limbonaut:
Starting with Godot 4.4, additional files are created for each script, and it happens that our routine executes for those files as well.

#skip-changelog

Screenshot 2025-06-27 at 16 24 16


jpnurmi (approved):
LGTM :+1:

PR #218, opened 2025-06-30
feat: Add 32-bit targets for Windows and Linux

limbonaut:
This PR adds 32 bit targets for Windows and Linux. To allow different architectures, the sentry-native compilation step needed to be reworked. It's a cleaner, separate, and more flexible routine now.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -118,55 +118,55 @@ env.Append(CPPPATH=["src/"])
 
 # Include sentry-native libs (static).
 if env["platform"] in ["linux", "macos", "windows"]:
-    env.Append(CPPDEFINES=["SENTRY_BUILD_STATIC", "NATIVE_SDK"])
-    env.Append(CPPPATH=["modules/sentry-native/include"])
-    env.Append(LIBPATH=["modules/sentry-native/install/lib/"])
+    # env.Append(CPPDEFINES=["SENTRY_BUILD_STATIC", "NATIVE_SDK"])
+    # env.Append(CPPPATH=["modules/sentry-native/include"])
+    # env.Append(LIBPATH=["modules/sentry-native/install/lib/"])
 
-    sn_targets = []
-    sn_sources = ["modules/sentry-native/src/"]
+    # sn_targets = []
+    # sn_sources = ["modules/sentry-native/src/"]
 
-    def add_target(lib_name):
-        env.Append(LIBS=[lib_name])
-        if env["platform"] == "windows":
-            sn_targets.append("modules/sentry-native/install/lib/" + lib_name + ".lib")
-            sn_targets.append("modules/sentry-native/install/lib/" + lib_name + ".pdb")
-        else:
-            sn_targets.append("modules/sentry-native/install/lib/lib" + lib_name + ".a")
+    # def add_target(lib_name):
+    #     env.Append(LIBS=[lib_name])
+    #     if env["platform"] == "windows":
+    #         sn_targets.append("modules/sentry-native/install/lib/" + lib_name + ".lib")
+    #         sn_targets.append("modules/sentry-native/install/lib/" + lib_name + ".pdb")
+    #     else:
+    #         sn_targets.append("modules/sentry-native/install/lib/lib" + lib_name + ".a")
 
-    add_target("sentry")
-    add_target("crashpad_client")
-    add_target("crashpad_handler_lib")
-    add_target("crashpad_minidump")
-    add_target("crashpad_snapshot")
-    add_target("crashpad_tools")
-    add_target("crashpad_util")
-    add_target("mini_chromium")
+    # add_target("sentry")
+    # add_target("crashpad_client")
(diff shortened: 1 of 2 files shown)


limbonaut:
@sentry review

src/sentry_sdk.cpp
```diff
@@ -274,7 +274,10 @@ SentrySDK::SentrySDK() {
 	// user extracts the distribution archive without preserving such permissions.
 	if (OS::get_singleton()->is_debug_build()) {
 		_fix_unix_executable_permissions("res://addons/sentry/bin/macos/crashpad_handler");
-		_fix_unix_executable_permissions("res://addons/sentry/bin/linux/crashpad_handler");
+		_fix_unix_executable_permissions("res://addons/sentry/bin/linux/x86_64/crashpad_handler");
+		_fix_unix_executable_permissions("res://addons/sentry/bin/linux/x86_32/crashpad_handler");
+		_fix_unix_executable_permissions("res://addons/sentry/bin/linux/arm64/crashpad_handler");
+		_fix_unix_executable_permissions("res://addons/sentry/bin/linux/arm32/crashpad_handler");
```
jpnurmi, line 297: can these exist? do we have linux arm builds?

jpnurmi:
Is there any test to ensure the newly built binaries load without problems? On a related note, the price tag for x86_32 is that the artifact size goes up from 99.8MB to 151MB.

src/sentry_sdk.cpp
limbonaut: We'll be adding those. I'm a bit concerned that I might forget about it. The routine won't break if those files don't exist.

limbonaut:
> Is there any test to ensure the newly built binaries load without problems? 

Not yet, maybe we could run unit tests on those architectures.

> On a related note, the price tag for x86_32 is that the artifact size goes up from 99.8MB to 151MB.

Yeah, it’s mostly the debug symbols. I’m still figuring out how to distribute them. For now, it’s all in one archive. But the GCC symbols are known for being quite large.

src/sentry_sdk.cpp
jpnurmi: Ok, no problem. I wasn't sure if it was intentional or slipped in.

jpnurmi:
If running the full test suite is problematic, some kind of minimal integration test that runs _something_ that requires the target libraries to be successfully loaded would be helpful, just to verify that the build artifacts are usable.

limbonaut:
Modified CI to run unit tests on x86_32 as well.

jpnurmi:
Nice! 👍 I see some errors in ubuntu x86_32 test output related to missing deps?

```
libfontconfig.so.1: cannot open shared object file: No such file or directory
ERROR: Can't open dynamic library: /home/runner/work/sentry-godot/sentry-godot/project/addons/sentry/bin/linux/x86_32/libsentry.linux.debug.x86_32.so. Error: libcurl.so.4: cannot open shared object file: No such file or directory.
   at: open_dynamic_library (drivers/unix/os_unix.cpp:896)
ERROR: Can't open GDExtension dynamic library: 'res://addons/sentry/sentry.gdextension'.
   at: open_library (core/extension/gdextension.cpp:702)
ERROR: Error loading extension: 'res://addons/sentry/sentry.gdextension'.
   at: load_extensions (core/extension/gdextension_manager.cpp:291)
Godot Engine v4.4.stable.official.4c311cbee - https://godotengine.org/
libudev.so.1: cannot open shared object file: No such file or directory
[...]
SCRIPT ERROR: Parse Error: Identifier "SentrySDK" not declared in the current scope.
          at: GDScript::reload (res://test/suites/test_sdk.gd:10)
[...]
debug> No test cases found, abort test run!
Exit code: 0
```

How come exit code 0 when it's full of errors? Do we have control over it?

limbonaut:
> How come exit code 0 when it's full of errors? Do we have control over it?

Those errors don't halt the engine, so it keeps going. And the testing framework is not capable of making sense why there are no tests. Hard to say what to do here...

UPDATED:

> I see some errors in ubuntu x86_32 test output related to missing deps?

Oh yeah, need the same deps as during builds I think.

.github/workflows/unit_tests.yml
```diff
@@ -5,26 +5,41 @@ on:
...
-            godot-suffix: macos.universal
+            arch: universal
     steps:
       - name: Checkout repo
         uses: actions/checkout@v4
         with:
           submodules: false # don't initialize submodules automatically
 
+      - name: Install Linux dependencies (32-bit)
+        if: runner == 'ubuntu-latest' && matrix.arch == 'x86_32'
```
jpnurmi:
```suggestion
        if: matrix.runner == 'ubuntu-latest' && matrix.arch == 'x86_32'
```

limbonaut: of course 😅 thanks

jpnurmi (approved):
Thanks for the tests! :pray:

PR #239, opened 2025-07-07
feat: Provide noop (stub) builds for unsupported platforms

### LINKED ISSUE #217: Provide no-op versions for unsupported platforms
To keep user projects from breaking on unsupported platforms and architectures, we should include special builds with SDK auto-disabled (no-op builds).

limbonaut:
- Provide noop builds for unsupported platforms and architectures.
- Noop builds provide API imlementation that doesn't perform any actions, which allows GDScript continue to compile and work as expected.
- This should allow users to export their projects on unsupported platforms using the same codebase without breaking.
- Resolves #217


limbonaut:
I made some additions to this PR: fixed wrong file paths in the manifest, and turned off debug symbols for noop builds to keep them slim – especially the web exports can suffer from the much bigger export sizes.

.github/workflows/build_gdextension.yml
```diff
@@ -128,14 +156,67 @@ jobs:
...
+            arch: arm64
+            debug-symbols: no
+
+          - name: 🍏 iOS (noop, simulator, release)
+            runner: macos-latest
+            platform: ios
+            target: template_release
+            arch: universal
+            debug-symbols: no
+            scons-flags: ios_simulator=yes
```
jpnurmi: does it help to keep them separate, or could the flags be passed in one go?
```
scons-flags: ios_simulator=yes debug_symbols=no
```

jpnurmi (approved):

.github/workflows/build_gdextension.yml
limbonaut: I guess it doesn't make much sense to keep it separate – consolidated the flags in `matrix.scons-flags`, and also moved macos deployment target spec into matrix flags.

.github/workflows/build_gdextension.yml
```diff
@@ -195,9 +272,20 @@ jobs:
...
 
+      - name: Prepare SCONSFLAGS
+        shell: bash
+        run: |
+          SCONSFLAGS="$SCONSFLAGS ${{ matrix.scons-flags }}"
+          # If not specified, default to debug_symbols=yes.
+          if [[ "$SCONSFLAGS" != *"debug_symbols="* ]]; then
+            SCONSFLAGS="$SCONSFLAGS debug_symbols=yes"
+          fi
+          echo "SCONSFLAGS=$SCONSFLAGS" >> $GITHUB_ENV
```
jpnurmi: it would not be many more lines to explicitly specify "yes" for the remaining matrix cases for easier discoverability. ;) however, if debug_symbols defaults to "no" in the build system, perhaps you could simply specify the "yes" cases in the matrix instead?

limbonaut: I see how this can be confusing. Alright, let's specify explicitly in each matrix entry. I think I prefer it to be explicit "yes" or "no" if we go this route.

jpnurmi (approved):

PR #252, opened 2025-07-17
ref: Move exported Sentry classes into `sentry` namespace

limbonaut:
- Move exported Sentry classes into the `sentry` namespace.
- Move files to more appropriate locations.
- This is needed for future Cocoa SDK integration – there is a lot of name clashing.

#skip-changelog as it's not a user-visible change


limbonaut:
I met with some issues integrating Cocoa SDK and came to a conclusion that it would be easier and cleaner to encapsulate our classes in a namespace than to make workarounds for these issues. This doesn't affect end-user in any way.

jpnurmi:
If it's not too much work, it could be more attractive to review step by step, one thing at a time. :upside_down_face:

limbonaut:
@jpnurmi This really is a single changeset. Since it moves classes into `sentry` namespace, they needed to go into `sentry` folder (I prefer folder structure to follow namespace logic if possible). The only "additional" thing I've added is moving `disabled` classes into their own `sentry/disabled` dir, just like sentry-native stuff lives in `sentry/native` and android in `sentry/android`. And because all these files moved, the includes needed to be adjusted across the project files. It's less complicated than it looks!

jpnurmi:
Ok, fair enough. Let's fix the conflict, and it's good to go. Sorry for leaving it aside. :)

jpnurmi (approved):

PR #258, opened 2025-07-19
fix: Allow configuration script to run even if SDK is initially disabled in project settings

### LINKED ISSUE #255: Configuration script is not executed if the SDK is disabled in project settings
The configuration script appears to be ineffective if the SDK is not initially enabled in the project settings. Given that a user could potentially set the SDK to enabled in the configuration script, it is necessary that we run the script and verify the value of `enabled`.

limbonaut:
- Fixes #255

### DIFF
--- a/src/sentry/sentry_sdk.cpp
+++ b/src/sentry/sentry_sdk.cpp
@@ -207,40 +207,6 @@ PackedStringArray SentrySDK::_get_global_attachments() {
 void SentrySDK::_initialize() {
 	sentry::util::print_debug("starting Sentry SDK version " + String(SENTRY_GODOT_SDK_VERSION));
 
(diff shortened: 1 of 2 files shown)


src/sentry/sentry_sdk.cpp
```diff
@@ -207,50 +207,58 @@ PackedStringArray SentrySDK::_get_global_attachments() {
 void SentrySDK::_initialize() {
 	sentry::util::print_debug("starting Sentry SDK version " + String(SENTRY_GODOT_SDK_VERSION));
 
+	// Initialize user if it wasn't set explicitly in the configuration script.
```
limbonaut, line 210: Moved to the top of the function – need to initialize user instance before we exit from the function.

jpnurmi (approved):

PR #262, opened 2025-07-23
feat: Add `disabled_in_editor_play` option

### LINKED ISSUE #236: Impossible to deactivate sentry for debug runtime build
### How do you use Sentry?

Sentry Saas

### Sentry SDK version

0.6.1

### How did you install the SDK?

GitHub release

### Godot version

v4.4.1

### Which platform?

macOS

### How to reproduce

I tried to disable SDK for editor and use override project settings, bu Sentry SDK is always loaded and run from editor

Image

### Expected result

It should be possible to not launch SDK for editor runtime build

### Actual result

_No response_

### LINKED ISSUE #251: Add an easy way to disable SDK during editor dev runs
We don't have a way to easily disable SDK during dev runs, which was pointed out in #236.

Currently it's not even possible through code due to a bug:
- #249 

Ideally, however, it should be a simple checkbox with no coding required. The proposal is to add a separate checkbox option for this: `disabled_in_editor_play`.

limbonaut:
Add an option to disable SDK when the project is run from within the Godot editor.

Screenshot 2025-07-23 at 10 13 02

- Resolves #251 
- Resolves #236

### DIFF
--- a/project/test/suites/test_options.gd
+++ b/project/test/suites/test_options.gd
@@ -14,6 +14,7 @@ func before_test() -> void:
 func test_bool_properties(property: String, test_parameters := [
 		["enabled"],
 		["disabled_in_editor"],
+		["disabled_in_editor_play"],
 		["debug"],
 		["attach_log"],
 		["attach_screenshot"],


limbonaut:
@jpnurmi We already have a flag called `enabled`, so we’re using `disabled_in_x` here to indicate an exception from the general rule. If `enabled` is `false`, we don’t consider `disabled_*` flags at all. Hence the name.

PR #284, opened 2025-08-01
fix: Prevent stack overflow in variant conversion

limbonaut:
This PR fixes a bug that could cause stack overflow when converting Variant in the NativeSDK. This can happen when there are circular references in nested arrays and/or dictionaries in GDScript.


jpnurmi (approved):
nit: 32 could be extracted as a constant

PR #305, opened 2025-08-07
feat: Add option to disable logging messages as breadcrumbs

### LINKED ISSUE #293: Add option to disable logging messages as breadcrumbs
Since `1.0.0-alpha.0`, we’ve started adding `Logger` messages (which are basically prints) as breadcrumbs. Based on the feedback we’ve received, it seems like some folks don’t always want to see these messages as breadcrumbs. Let's add an option to turn off logging of non-error messages.

limbonaut:
This PR adds option `logger/messages_as_breadcrumbs` which is enabled by default. This allows disabling messages as breadcrumbs by a simple toggle. Since some projects `print()` heavily into console, this option should be quite useful.

- Resolves #293 

Screenshot 2025-08-07 at 19 15 49

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Intoduce `SentryTimestamp` class to improve timestamp handling ([#286](https://github.com/getsentry/sentry-godot/pull/286))
 - Support for iOS & macOS using Sentry Cocoa SDK integration ([#266](https://github.com/getsentry/sentry-godot/pull/266))
+- Add option to disable logging messages as breadcrumbs ([#305](https://github.com/getsentry/sentry-godot/pull/305))
 
 ### Improvements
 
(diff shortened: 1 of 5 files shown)


jpnurmi:
The Native SDK's [Qt integration](https://github.com/getsentry/sentry-native/blob/master/src/integrations/sentry_integration_qt.cpp) does something similar and forwards all Qt logging output to Sentry as breadcrumbs. Personally, I find it quite noisy and annoying, which makes me wonder if the feature would be better off as opt-in... How do other SDKs handle this? :thinking:

mujacica (approved):

mujacica:
@jpnurmi I think this is going to change with the upcoming Sentry Logs and logger changes. That should give us a possibility to configure these things in much more detail.

limbonaut:
This only affects projects that add a lot of `print()` statements in code. By default, this adds close to zero noise. Since it's project-dependent how much noise it adds, I think leaving it ON by default is good for discoverability.

PR #326, opened 2025-08-26
wip: Add integration tests

limbonaut:
This is a mega-branch of different PRs I'm working on in tandem here. I'm splitting work into separate PRs as the code gets ready. They all share connection to integration testing, and reevaluation of our API in the process.

The JSONAssert has been moved to its own repository: https://github.com/limbonaut/gdunit-json-assert
I propose we adopt it in getsentry.

#skip-changelog for now

PR #329, opened 2025-08-28
feat: Add SentryEvent.to_json()

limbonaut:
Add `SentryEvent.to_json()` in public API just like in other SDKs. Extracted from #326. 
Tests are in #326.

### DIFF
--- a/src/sentry/native/native_event.cpp
+++ b/src/sentry/native/native_event.cpp
@@ -215,6 +215,10 @@ bool NativeEvent::is_crash() const {
 	return _is_crash;
 }
 
+String NativeEvent::to_json() const {
+	return String::utf8(sentry_value_to_json(native_event));
+}
+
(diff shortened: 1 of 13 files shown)


tustanivsky (approved):

src/sentry/native/native_event.cpp
```diff
@@ -215,6 +215,10 @@ bool NativeEvent::is_crash() const {
 	return _is_crash;
 }
 
+String NativeEvent::to_json() const {
+	return String::utf8(sentry_value_to_json(native_event));
```
jpnurmi: possible memory leak: the return value of `sentry_value_to_json` needs to be freed?

limbonaut: Of course, thank you! 🙏

jpnurmi (approved):

PR #331, opened 2025-08-28
chore: Don't add breadcrumb data and category when empty

limbonaut:


limbonaut:
Closing in favor of #332

Issue #163, opened 2025-04-17
Support compiling as Godot Engine module (as part of the engine)


limbonaut:
Currently, we only support compilation as GDExtension. Add support for compiling as engine module, in other words, as a part of Godot Engine.

Please 👍 if this is something you'd like to have support for.

PR #332, opened 2025-08-29
feat: Introduce SentryBreadcrumb class

### LINKED ISSUE #338: Introduce `SentryBreadcrumb` class
Introduce `SentryBreadcrumb` class for cleaner API and to be able to support `before_breadcrumb` hook in future. 

Currently,`add_breadcrumb()` method accepted 5 parameters (3 of which were strings), making it confusing to use. The new approach would use a dedicated `SentryBreadcrumb` class:

```gdscript
var crumb := SentryBreadcrumb.create("Something happened")
crumb.type = "info"
crumb.set_data({"some": "data"})
SentrySDK.add_breadcrumb(crumb)
```

This would provide better type safety, improved readability, and enables future support for the `before_breadcrumb` callback. We need to prioritize this for 1.0.0 as it's a breaking change in a widely used API.

limbonaut:
This PR adds a new breadcrumb API for a cleaner, more intuitive interface. Previously, `add_breadcrumb()` method accepted 5 parameters (3 of which were strings), making it confusing to use. The new approach uses a dedicated `SentryBreadcrumb` class:

```gdscript
var crumb := SentryBreadcrumb.create("Something happened")
crumb.type = "info"
crumb.set_data({"some": "data"})
SentrySDK.add_breadcrumb(crumb)
```

For simple breadcrumbs, you can use a one-liner:
```gdscript
SentrySDK.add_breadcrumb(SentryBreadcrumb.create("Something happened"))
```

This change provides better type safety, improved readability, and enables future support for the `before_breadcrumb` callback.

- Resolves #338

### DIFF
--- a/android_lib/src/main/java/io/sentry/godotplugin/SentryAndroidGodotPlugin.kt
+++ b/android_lib/src/main/java/io/sentry/godotplugin/SentryAndroidGodotPlugin.kt
@@ -110,7 +110,7 @@ class SentryAndroidGodotPlugin(godot: Godot) : GodotPlugin(godot) {
         }
 
         var handle = Random.nextInt()
-        while(breadcrumbsMap.containsKey(handle)) {
+        while (breadcrumbsMap.containsKey(handle)) {
             handle = Random.nextInt()
         }
 
--- a/android_lib/src/main/java/io/sentry/godotplugin/SentryAndroidGodotPlugin.kt
+++ b/android_lib/src/main/java/io/sentry/godotplugin/SentryAndroidGodotPlugin.kt
@@ -43,6 +43,12 @@ class SentryAndroidGodotPlugin(godot: Godot) : GodotPlugin(godot) {
         }
     }
 
+    private val breadcrumbsByHandle = object : ThreadLocal<MutableMap<Int, Breadcrumb>>() {
+        override fun initialValue(): MutableMap<Int, Breadcrumb> {
+            return mutableMapOf()
+        }
+    }
+
     private fun getEvent(eventHandle: Int): SentryEvent? {
         val event: SentryEvent? = eventsByHandle.get()?.get(eventHandle)
         if (event == null) {
@@ -57,6 +63,14 @@ class SentryAndroidGodotPlugin(godot: Godot) : GodotPlugin(godot) {
             Log.e(TAG, "Internal Error -- SentryException not found: $exceptionHandle")
         }
         return exception
+    }
+
+    private fun getBreadcrumb(breadcrumbHandle: Int): Breadcrumb? {
+        var crumb: Breadcrumb? = breadcrumbsByHandle.get()?.get(breadcrumbHandle)
+        if (crumb == null) {
+            Log.e(TAG, "Internal Error -- Breadcrumb not found: $breadcrumbHandle")
+        }
+        return crumb
     }
 
     private fun registerEvent(event: SentryEvent): Int {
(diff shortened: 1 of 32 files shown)


src/sentry/sentry_breadcrumb.h
```diff
@@ -0,0 +1,44 @@
+#ifndef SENTRY_BREADCRUMB_H
+#define SENTRY_BREADCRUMB_H
+
+#include "sentry/level.h"
+#include "sentry/sentry_timestamp.h"
+
+#include <godot_cpp/classes/ref_counted.hpp>
+
+using namespace godot;
```
jpnurmi, line 9: potentially for the future / to be addressed separately: not sure if it has been discussed before, but wouldn't it be better to leave this out of public headers?

src/sentry/native/native_breadcrumb.cpp
```diff
@@ -0,0 +1,81 @@
...
+}
+
+void NativeBreadcrumb::set_data(const Dictionary &p_data) {
+	sentry_value_t native_data = sentry::native::variant_to_sentry_value(p_data);
+	sentry_value_set_by_key(native_crumb, "data", native_data);
+}
+
+Ref<SentryTimestamp> NativeBreadcrumb::get_timestamp() {
+	sentry_value_t value = sentry_value_get_by_key(native_crumb, "timestamp");
+	sentry::util::print_debug("breadcrumb timestamp: ", String(sentry_value_as_string(value)));
```
jpnurmi: is this intentional or could be cleaned up?

src/sentry/disabled/disabled_breadcrumb.h
```diff
@@ -0,0 +1,42 @@
...
+
+	virtual void set_level(sentry::Level p_level) override { level = p_level; }
+	virtual sentry::Level get_level() const override { return level; }
+
+	virtual void set_type(const String &p_type) override { type = p_type; }
+	virtual String get_type() const override { return type; }
+
+	virtual void set_data(const Dictionary &p_data) override { data = p_data; }
+
+	virtual Ref<SentryTimestamp> get_timestamp() override { return memnew(SentryTimestamp); }
```
jpnurmi, line 37: does this tend to get called multiple times? would it make sense to allocate the timestamp once as a member if an instance is needed, or return a null ref?

android_lib/src/main/java/io/sentry/godotplugin/SentryAndroidGodotPlugin.kt
```diff
@@ -89,6 +103,21 @@ class SentryAndroidGodotPlugin(godot: Godot) : GodotPlugin(godot) {
...
     }
 
+    private fun registerBreadcrumb(crumb: Breadcrumb): Int {
+        val breadcrumbsMap = breadcrumbsByHandle.get() ?: run {
+            Log.e(TAG, "Internal Error -- breadcrumbsByHandle is null")
+            return 0
+        }
+
+        var handle = Random.nextInt()
+        while(breadcrumbsMap.containsKey(handle)) {
```
jpnurmi: nit
```suggestion
        while (breadcrumbsMap.containsKey(handle)) {
```

src/sentry/sentry_breadcrumb.h
limbonaut: In Godot Engine code base, all the classes live in global namespace. Because we consider supporting compilation as an engine module, I treat this namespace as global as well, at least where it comes to our own class headers that we expose to Godot like SentryBreadcrumb. See #163.

src/sentry/native/native_breadcrumb.cpp
limbonaut: Not at all intentional, thanks!

src/sentry/disabled/disabled_breadcrumb.h
limbonaut: We don't call it at all, only user would do that possibly ~~within before_breadcrumb~~ (correction: not in Disabled backend). I don't think it's worth optimizing. The only reason why we return a non-null is not to break script code.

For example:
```gdscript
var crumb := SentryBreadcrumb.create()
print(crumb.get_timestamp()) #  null access
do_more()
```
The do_more() part would break in noop builds, while being perfectly fine in normal builds.

PR #334, opened 2025-09-02
feat: Support local variables on Android

limbonaut:
#skip-changelog for now


android_lib/src/main/java/io/sentry/godotplugin/SentryAndroidGodotPlugin.kt
```diff
@@ -442,6 +442,11 @@ class SentryAndroidGodotPlugin(godot: Godot) : GodotPlugin(godot) {
                 preContext = (frameData["pre_context"] as? Array<*>)?.map { it as String }
                 postContext = (frameData["post_context"] as? Array<*>)?.map { it as String }
             }
+
+            if (frameData.containsKey("vars")) {
+                val variables = frameData["vars"] as? Dictionary
```
tustanivsky: nit: maybe assign `vars` directly instead of using an intermediate variable

limbonaut: Updated in a1042e9

Issue #337, opened 2025-09-02
Add tests for UTF-8 retention across API calls that deal with strings


limbonaut:
Seeing that we often convert strings between different frameworks, we could use basic tests for UTF-8 retention.

limbonaut:
Partially addressed in #345 and the rest is coming in #326, so I'm closing this.

PR #345, opened 2025-09-03
fix: UTF-8 retention problems with native SentryEvent properties

### LINKED ISSUE #336: UTF-8 not preserved in some API calls on Linux and Windows
In some places such as set_tag() and get_tag(), get_tag() would not preserve UTF-8 encoding, returning a different string. This does affect mostly getters, not setters.

When we take C strings from native SDK and convert to Godot strings, we need to use `String::utf8` where those C strings realistically can contain UTF-8 encoded data.

limbonaut:
This PR fixes issues with retaining UTF-8 encoding in native SentryEvent implementation, discovered while working on tests in #326. The problem is that getters should be using String::utf8() where we reasonably expect UTF-8. Extracted from #326.

- Fixes #336
- Partially implements #337, more are coming in #326

### DIFF
--- a/project/test/suites/test_event.gd
+++ b/project/test/suites/test_event.gd
@@ -2,18 +2,26 @@ extends GdUnitTestSuite
 ## Basic tests for the SentryEvent class.
 
 
+## Test string properties accessors and UTF-8 encoding preservation.
+@warning_ignore("unused_parameter")
+func test_string_properties_and_utf8(property: String, test_parameters := [
+		["message"],
+		["logger"],
+		["release"],
+		["dist"],
+		["environment"],
+]) -> void:
+	var event := SentrySDK.create_event()
+	event.set(property, "Hello, World!")
+	assert_str(event.get(property)).is_equal("Hello, World!")
+	event.set(property, "Hello 世界! 👋")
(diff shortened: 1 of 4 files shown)


project/test/suites/test_event.gd
```diff
@@ -2,18 +2,26 @@ extends GdUnitTestSuite
...
+		["release"],
+		["dist"],
+		["environment"],
+]) -> void:
 	var event := SentrySDK.create_event()
-	assert_str(event.id).is_not_empty()
+	event.set(property, "Hello, World!")
+	assert_str(event.get(property)).is_equal("Hello, World!")
+	event.set(property, "Hello 世界! 👋")
+	assert_str(event.get(property)).is_equal("Hello 世界! 👋")
```
limbonaut, line 19: This replaces multiple tests removed below.

limbonaut:
@jpnurmi Yes, we convert Godot types to internal SDK types recursively, but not the other way around (at least not yet). We’ll be testing it in one of the future PRs. Here’s a work-in-progress glimpse: https://github.com/getsentry/sentry-godot/pull/326/commits/c5969bac48330b4a1afcc25f002ea3fdea9678ec

Issue #354, opened 2025-09-11
Unable to set up code mapping for GDScript files


limbonaut:
It was reported that "Set up Code Mapping" dialog fails to map GDScript files to GitHub repository, with "We can't find the file path for ..." error.
When we log script errors, we leave file paths untouched in GDScript stack traces. These paths contain Godot-specific prefixes such as "res://". This prefix means that this file path is a file within the Godot project (typically, within the pck file). This is the likely culprit.

limbonaut:
This isn't necessary, because you can set up project root mapping in the sentry.io settings.

PR #355, opened 2025-09-11
fix: Strip "res://" prefix in GDScript stack trace files

limbonaut:
- #354 

#skip-changelog for now


limbonaut:
It seems like this isn’t necessary, since you can set up project root mapping in settings.

PR #358, opened 2025-09-12
chore: Fix flaky timestamp test case

limbonaut:
Few times I hit this issue:
Screenshot 2025-09-04 at 17 34 46

This PR should allow more tolerance for this test.

#skip-changelog

### DIFF
--- a/project/test/suites/test_breadcrumb.gd
+++ b/project/test/suites/test_breadcrumb.gd
@@ -49,7 +49,7 @@ func test_breadcrumb_default_values() -> void:
 
 
 func test_breadcrumb_timestamp_is_set_automatically() -> void:
-	const time_tolerance: float = 0.05  # 50 ms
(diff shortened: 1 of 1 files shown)


project/test/suites/test_breadcrumb.gd
```diff
@@ -49,13 +49,15 @@ func test_breadcrumb_default_values() -> void:
 
 
 func test_breadcrumb_timestamp_is_set_automatically() -> void:
-	var time_before: float = Time.get_unix_time_from_system()
+	const time_tolerance: float = 0.05  # 50 ms
```
jpnurmi: Wouldn't something like 1ms be enough to account for floating-point precision issues?

limbonaut: Sure, let's make it tighter.

jpnurmi (approved):

PR #379, opened 2025-09-22
chore: Add "Run tests" button instead of running tests on start in demo

limbonaut:
This PR adds a separate "Run tests" button in a "Test" tab instead of running tests on project start. Running these tests on start, like it was previously, interferes with crash testing and before_send logic in tests.


limbonaut:
This also fixes an error when running the demo project on mobile straight from release package.

PR #395, opened 2025-09-29
chore: Remove get_user() call from demo

limbonaut:
Remove `SentrySDK.get_user()` call from the demo project, since we removed it from API in #392.


limbonaut:
I had to refactor user interface logic in the demo, becaus we no longer expose get_user().

jpnurmi (approved):

PR #407, opened 2025-10-06
chore: Manually upgrade gdUnit4 to v6.0.0

limbonaut:
The upgrade to gdUnit4 testing framework v6.0.0 requires removing compatibility patch, and additionally updating JSONAssert submodule due to API changes.


CHANGELOG.md
```diff
@@ -43,6 +43,8 @@
   - [changelog](https://github.com/getsentry/sentry-native/blob/master/CHANGELOG.md#0112)
   - [diff](https://github.com/getsentry/sentry-native/compare/0.10.1...0.11.2)
 
+- Manually upgrade gdUnit4 to v6.0.0 ([#407](https://github.com/getsentry/sentry-godot/pull/407))
+
```
vaind: THis is an internal dependency, right? no changelog item in that case.
```suggestion
```

vaind (approved):

CHANGELOG.md
limbonaut: The bots [usually add](https://github.com/getsentry/sentry-godot/pull/377) an item for it.

vaind: [That can be disabled](https://github.com/getsentry/sentry-godot/pull/408). If it's an internal dependency, it shouldn't have a changelog item, just like any other internal change - it's just spam for consumers

project/test/isolated/test_limit_throttling.gd
```diff
@@ -27,13 +27,14 @@ func _before_send(_ev: SentryEvent) -> SentryEvent:
...
+
 	push_error("dummy-error")
 	push_error("dummy-error")
 	push_error("dummy-error")
-	await assert_signal(monitor).is_emitted("callback_processed")
-	await assert_signal(monitor).is_emitted("callback_processed")
-	await get_tree().create_timer(0.1).timeout
+
+	await assert_signal(self).is_emitted("callback_processed")
+	await get_tree().process_frame  # allow all events to process
```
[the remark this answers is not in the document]
limbonaut: There is some issue with this approach and gdUnit4 6.0.0. Can't say why without digging into the internals of gdUnit. But the revised test itself is correct. We are verifying if the callback is called, and then comparing the counter to expected value. So it's fine at it is.

PR #414, opened 2025-10-11
chore: Refactor logging & avoid logger errors caused by SDK debug printing

limbonaut:
Refactors engine logging in preparation for structured logs (#409).

This PR:
- Relocates `SentryLogger` to the `logging` namespace and renames it to `SentryGodotLogger`.
  - This is internally used engine logger -- added detailed description in comments.
  - The renaming is needed as we're introducing structured logs and need to expose `SentrySDK.logger` interface in the API, which will take over the `SentryLogger` class name.
- Moves `print.h` to the `logging` namespace, as it belongs there.
- **Improvement**: Implements detection to identify when we're inside message logging, preventing print operations through the Godot logger to avoid runtime errors.

### DIFF
--- a/src/sentry/cocoa/cocoa_event.mm
+++ b/src/sentry/cocoa/cocoa_event.mm
@@ -1,7 +1,6 @@
 #include "cocoa_event.h"
 
 #include "cocoa_util.h"
-#include "sentry/logging/print.h"
 
 #include <godot_cpp/classes/os.hpp>
 


src/sentry/cocoa/cocoa_event.mm
```diff
@@ -1,7 +1,7 @@
 #include "cocoa_event.h"
 
 #include "cocoa_util.h"
-#include "sentry/util/print.h"
+#include "sentry/logging/print.h"
```
jpnurmi: possibly unused if nothing had to be changed regardless of the namespace change?

limbonaut:
> Did you consider some completely different name? Maybe something like "console" could work? 

Yeah, I know. It's called [Logging](https://docs.godotengine.org/en/stable/tutorials/scripting/logging.html) in Godot and console doesn't really fit either. In the end, I decide to call the Godot subsystem "logging" and the sentry part "structured logs". But it's the class names right? The base class for Godot's is `Logger` and it's hard to figure something that would fit here and is not Logger. My other candidates are `SentryEngineLogger` and `SomethingLoggerIntegration`. 

UPDATE:
We also have `logger_*` options that affect engine logger, so keeping "Logger" in class name is probably a good idea.

src/sentry/logging/sentry_godot_logger.cpp
```diff
@@ -403,18 +408,21 @@ void SentryLogger::_log_message(const String &p_message, bool p_error) {
 		}
 	}
 
-	Ref<SentryBreadcrumb> crumb = SentryBreadcrumb::create(processed_message);
-	crumb->set_category("log");
-	crumb->set_level(p_error ? sentry::Level::LEVEL_ERROR : sentry::Level::LEVEL_INFO);
-	crumb->set_type("debug");
-	SentrySDK::get_singleton()->add_breadcrumb(crumb);
+	if (as_breadcrumb) {
```
jpnurmi, line 411: left-over? cannot be false because there's `if (!as_breadcrumb) return` above

limbonaut: Aligning with #409 as this section will be expanded to also process `as_log`.

PR #416, opened 2025-10-14
feat: Set app hang timeout to 5s on Apple platforms

limbonaut:
Increase app hang timeout on Apple platforms to 5 seconds since it can give false positives with 2s.

### DIFF
--- a/.github/actions/prepare-testing/action.yml
+++ b/.github/actions/prepare-testing/action.yml
@@ -32,13 +32,18 @@ runs:
     - name: Set up Godot
       shell: bash
       env:
-        # Set pre-release version status here (e.g. "beta1");
-        # for release version, set to "stable".
-        RELEASE_STATUS: stable
+        # Set Godot version override here (e.g. "4.5.1-stable").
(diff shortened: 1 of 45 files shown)


jpnurmi:
Out of curiosity, what kind of platforms/devices/environments can give false positives with 2s, and during which phase of the game lifecycle? 5s sounds like quite a long time for a game not to respond, except perhaps during startup? In the end, both 2s and 5s are arbitrary numbers, but how do they compare to other platforms and SDKs?

limbonaut:
I noticed this behavior on macOS during occasional project startup: the app hanging event is sent, and it looks as a normal load otherwise. I believe you can also trigger it if you load heavy resources synchronously on the main thread. Perhaps, we could completely disable app hanging integration during debug runs. I'm not sure how this issue affects the average release export out there. But in this case, I believe it is spam.

@bitsandfoxes mentioned that Unity also has 5s as the default. I don’t think Native has such integration, and I haven’t seen this event triggering on Android.

PR #417, opened 2025-10-14
chore: Fix flaky throttling test

limbonaut:
Add a bit more time for the throttling window to expire because we process it once per frame and it can take a little longer than the specified window.

### DIFF
--- a/project/test/isolated/test_limit_throttling.gd
+++ b/project/test/isolated/test_limit_throttling.gd
@@ -38,7 +38,7 @@ func test_throttling_limits() -> void:
 	assert_int(_num_events).is_equal(2)
 
 	# Wait for throttling window to expire.
-	await get_tree().create_timer(1.0).timeout
+	await get_tree().create_timer(1.1).timeout
 
 	push_error("dummy-error")
 	push_error("dummy-error")


project/test/isolated/test_limit_throttling.gd
```diff
@@ -38,7 +38,7 @@ func test_throttling_limits() -> void:
 	assert_int(_num_events).is_equal(2)
 
 	# Wait for throttling window to expire.
-	await get_tree().create_timer(1.0).timeout
+	await get_tree().create_timer(1.1).timeout
```
tustanivsky, line 41: Just curious - do we have any data on the typical throttling window delays that justify using 1.1?

limbonaut: We recalculate throttling window data [once per frame](https://github.com/getsentry/sentry-godot/blob/3e54770c69fec41f49104a5990c40de072cefdb1/src/sentry/logging/sentry_godot_logger.cpp#L252-L269). It should be 1000 ms + whatever the frame takes. 

This test doesn’t need to be overly strict. We just need to make sure it resets at some point after the configured window time.

PR #422, opened 2025-10-17
feat: User feedback form

### LINKED ISSUE #170: User Feedback Widget
Sentry has a User Feedback feature that is able to show the feedback collected from a user following an error, or simply initiated by a user (without the need to be connected to an error/event).
* https://docs.sentry.io/product/user-feedback/

The feedback collection allows attaching trace information as well as Session Replay. In addition to other context like tags.
In terms of SDK support, in JS we added a web widget, that's easy to add to the screen:
* https://docs.sentry.io/platforms/javascript/user-feedback/

The goal is to add a similar capability to mobile SDKs.

The SDKs have a new method that represents this new protocol: `CaptureFeedback`. The old one should be deprecated in favor of the new one 

See:
* https://develop.sentry.dev/application-architecture/feedback-architecture/
* https://github.com/getsentry/sentry-cocoa/issues/4187

The widget has been added to the iOS SDK with quite a few features already. And is also being added to the other major SDKs:
* https://github.com/getsentry/sentry-react-native/issues/4302
* https://github.com/getsentry/sentry-cocoa/issues/4270
* https://github.com/getsentry/sentry-java/issues/3613
* https://github.com/getsentry/sentry-unreal/issues/893
* * https://github.com/getsentry/sentry-dotnet/issues/3964
* https://github.com/getsentry/sentry-dart/blob/7954fb356f96f17127c41230af433bc9ce01233c/flutter/lib/src/feedback/sentry_feedback_widget.dart#L55

Some ideas are new and are not aligned across all SDKs, for example:
* https://github.com/getsentry/sentry-dart/issues/2541
But ideally we have a similar behavior/look/feel as much as that makes sense. Things should be idiomatic (e.g: swiping down in the modal removes the feedback widget for example)

limbonaut:
This PR adds customizable User Feedback UI form. This is a reference implementation in GDScript, that developers can use in their projects for feedback submission. It supports some customization and theming, and can be used as an example to kickstart custom implementations.

User feedback form is located in `addons/sentry/user_feedback` folder.

The feedback form is customizable in appearance and behavior:
- Appearance can be customized by copying the whole folder into the project, and adjusting UI controls directly in the scene files, and/or by adjusting the associated UI theme file. The reference theme file is included.
- It automatically handles message validation, character limits, and feedback submission.

There are 3 sets of files:
- `user_feedback_form.{tscn,gd}`: Feedback form -- a flexible control for integration into existing UI.
  - _Doesn't require the other files to function._
- `user_feedback_gui.{tscn,gd}`: Ready to use integration -- drag & drop into project's UI scene and call show() when needed.
  - _Serves as an example of how to integrate feedback form into a project._
- `sentry_theme.tres` Reference UI theme file (using light Sentry color scheme).

Additionally, the form can be customized via exported properties:
- `logo_visible` -- if `true`, shows Sentry logo in the top-right corner (see screenshot).
- `name_visible` -- shows/hides name input field.
- `email_visible` -- shows/hides email input field.
- `minimum_words` -- the number of words required before "Submit" button is enabled.

The default integration (`user_feedback_gui`) also handles auto-scaling for different resolutions.

- Resolves #170
- Docs https://github.com/getsentry/sentry-docs/pull/15304

**Screenshot:**

Screenshot 2025-10-18 at 14 17 58

**Auto-scaling demonstration:**

I disabled window stretch to simulate different resolutions (1 unit in scene corresponds to 1 px in the window).

https://github.com/user-attachments/assets/450667a0-4118-40bd-a49a-b471d7ac2ea6

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,13 +5,15 @@
 ### Features
 
 - Add user feedback API for collecting and sending user feedback to Sentry ([#418](https://github.com/getsentry/sentry-godot/pull/418))
-- Access event exception values in `before_send` handler ([#415](https://github.com/getsentry/sentry-godot/pull/415))
 - Add customizable User Feedback form that can be used for feedback submission, and as an example for custom implementations ([#422](https://github.com/getsentry/sentry-godot/pull/422))
+- Access event exception values in `before_send` handler ([#415](https://github.com/getsentry/sentry-godot/pull/415))
+- Add support for Structured Logging ([#409](https://github.com/getsentry/sentry-godot/pull/409))
 
 ### Improvements
 
 - Detect when we're inside message logging to prevent SDK print operations through the Godot logger which cause runtime errors. ([#414](https://github.com/getsentry/sentry-godot/pull/414))
 - Relax throttling limits on app startup ([#423](https://github.com/getsentry/sentry-godot/pull/423))
+- Set app hang timeout to 5s on Apple platforms ([#416](https://github.com/getsentry/sentry-godot/pull/416))
 
 ### Dependencies
 
@@ -21,6 +23,9 @@
 - Bump Cocoa SDK from v8.56.2 to v8.57.0 ([#419](https://github.com/getsentry/sentry-godot/pull/419))
   - [changelog](https://github.com/getsentry/sentry-cocoa/blob/main/CHANGELOG.md#8570)
   - [diff](https://github.com/getsentry/sentry-cocoa/compare/8.56.2...8.57.0)
+- Bump Sentry Android from v8.23.0 to v8.24.0 ([#424](https://github.com/getsentry/sentry-godot/pull/424))
+  - [changelog](https://github.com/getsentry/sentry-java/blob/main/CHANGELOG.md#8240)
+  - [diff](https://github.com/getsentry/sentry-java/compare/8.23.0...8.24.0)
 
 ## 1.0.0
 
--- a/android_lib/build.gradle.kts
+++ b/android_lib/build.gradle.kts
@@ -38,7 +38,7 @@ dependencies {
     testImplementation("junit:junit:4.13.2")
 
     // NOTE: All dependencies below must be also updated in sentry_editor_export_plugin.cpp.
-    implementation("io.sentry:sentry-android:8.23.0")
+    implementation("io.sentry:sentry-android:8.24.0")
 }
 
 val copyDebugAarToProject by tasks.registering(Copy::class) {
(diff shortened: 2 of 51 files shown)


project/addons/sentry/user_feedback/user_feedback_form.gd
```diff
@@ -0,0 +1,106 @@
...
+func _ready() -> void:
+	_update_controls()
+	_on_message_edit_text_changed()
+
+
+func _update_controls() -> void:
+	if is_node_ready():
+		%Logo.visible = show_logo
+		%EmailSection.visible = enable_email_input
+		%NameSection.visible = enable_name_input
```
jpnurmi: if the godot api uses "visible", could we align with that instead of "show" and "enable"? especially the latter sounds potentially misleading

jpnurmi (commented):
Nice! I might steal the layout order with a mandatory message at the top, and an optional name and email below. Makes more sense that way! 🙂 

By the way, are the `.tscn` and `.tres` files WYSIWYG-editable with a form designer or some such?

limbonaut:
> By the way, are the `.tscn` and `.tres` files WYSIWYG-editable with a form designer or some such?

Oh yeah, those are not manually typed! `tscn` stands for "text scene" and `tres` for "text resource" (Git-friendly serialization format), as opposed to scn and res, which are binary. `tscn` stores a scene that is edited directly in the Godot editor. `tres` is the same format as `tscn`, just used for some other `Resource`-derived types. Most of the builtin resources that serialize to tres have some kind of dedicated editor space in Godot, but not always. Here are some screenshots to illustrate it:

project/addons/sentry/user_feedback/user_feedback_form.gd
limbonaut: I found all three patterns used in Godot properties; neither is more prominent, it seems. I think we should stick with one of those.

limbonaut: Oh yeah, there is also "x_enabled" 😂



What do you think about going with "x_enabled" for all three props @jpnurmi ?

jpnurmi: To me, "visible" would be a clear winner. :upside_down_face: 

- "show": sounds like a function, not a property
- "enable(d)": in the context of a UI control, the opposite of being disabled i.e. non-interactive

limbonaut: I see your point 🤔 Let's go with your suggestion!

project/addons/sentry/user_feedback/user_feedback_form.gd
```diff
@@ -0,0 +1,106 @@
...
+	_message_edit.text = ""
+
+
+func _on_message_edit_text_changed() -> void:
+	var message: String = _message_edit.text
+	if message.length() > 4096:
+		var col: int = _message_edit.get_caret_column()
+		message = message.substr(0, 4096)
+		_message_edit.text = message
+		_message_edit.set_caret_column(col)
```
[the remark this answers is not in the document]
limbonaut: The caret will be clamped to the valid range. No issue.

PR #433, opened 2025-10-31
test: Mobile test runner

limbonaut:
This PR:
- Introduces a custom test runner, that can be executed directly from code, and bumps gdUnit4 to required commit.
- Updates CI to use the new test runner, since it improves readability in GH logs with lots of output.
- To initiate tests, the demo project expects the following environment variables to be set: 
  ```bash
  SENTRY_TEST=1
  SENTRY_TEST_INCLUDE="res://test/suites/"   # ";"-separated list of files and/or directories
  ```
  - Additionally, these variables can be set via Android extent extras using the same names.
- Adds a script to run tests on Android on a locally connected device.
- Updates CI to run tests on Android emulator.
- Adds a tool button to run tests straight from the demo project for convenience.
- Adjusts tests for Android platform specifics, and removes mobile tests, which are no longer needed.
- Depends on #436 

#skip-changelog

PR #436, opened 2025-11-04
chore: Use threads interface on Android for script stack traces

limbonaut:
Use threads interface for GDScript stack traces on Android, similar to how we do it on Native and Cocoa. Also simplifies exception-related code.

#skip-changelog


limbonaut:
I've ran tests in #433 and it's all good.

PR #439, opened 2025-11-06
fix: Crash on exit on Android

limbonaut:
This PR addresses a bug that occasionally caused crashes when the app exited on Android. It avoids calling into the bridge layer after it’s likely gone and hardens the internals of the `AndroidSDK` to prevent similar issues. The issue was discovered while working on #433.

https://sentry-sdks.sentry.io/issues/7003155763/events/2f6a9daafe954d302e4c3f1daf2d5902/

### DIFF
--- a/src/sentry/android/android_sdk.cpp
+++ b/src/sentry/android/android_sdk.cpp
@@ -77,26 +77,31 @@ void SentryAndroidBeforeSendLogHandler::_bind_methods() {
 // *** AndroidSDK
 
 void AndroidSDK::set_context(const String &p_key, const Dictionary &p_value) {
+	Object *android_plugin = _get_android_plugin();
 	ERR_FAIL_NULL(android_plugin);
 	android_plugin->call(ANDROID_SN(setContext), p_key, sanitize_variant(p_value));
 }
 
 void AndroidSDK::remove_context(const String &p_key) {
+	Object *android_plugin = _get_android_plugin();
 	ERR_FAIL_NULL(android_plugin);
 	android_plugin->call(ANDROID_SN(removeContext), p_key);
 }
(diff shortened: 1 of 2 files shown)


src/sentry/android/android_sdk.cpp
```diff
@@ -274,10 +294,6 @@ AndroidSDK::AndroidSDK() {
 }
 
 AndroidSDK::~AndroidSDK() {
-	if (is_enabled()) {
-		close();
-	}
```
jpnurmi, line 279: not sure if it matters that this deviates from the native & cocoa implementations that still call close(), but can the instance be alive under certain circumstances, and should it close if so?

src/sentry/android/android_sdk.cpp
```diff
@@ -274,10 +294,6 @@ AndroidSDK::AndroidSDK() {
...
-
 	AndroidStringNames::destroy_singleton();
```
jpnurmi, line 297: should `android_plugin_instance_id` be reset somewhere?

src/sentry/android/android_sdk.cpp
limbonaut: I'm not even sure if we should close() at all. There's probably some contingency set up to flush on application exit anyway.

src/sentry/android/android_sdk.cpp
limbonaut: Not really, the plugin is a unique instance singleton that gets managed by Godot. ObjectDB::get_instance(instance_id) returns null for objects that were freed. So even if we retain a number, there is no risk involved.

jpnurmi (approved):

PR #462, opened 2025-12-10
test: iOS end-to-end in CI

limbonaut:
This PR adds iOS end-to-end testing in CI. In order to run such tests in SauceLabs, we need to build iOS application and code sign it with "Development" certificate and profile, so it can be launched on a real device.

### DIFF
--- a/.github/workflows/test_integration.yml
+++ b/.github/workflows/test_integration.yml
@@ -103,7 +103,6 @@ jobs:
         run: |
           mkdir -p "$(dirname "${{github.workspace}}/${{matrix.test-executable}}")"
 
-          # TODO: Export fails because Godot attempts to build the Xcode project, which requires a certificate and provisioning profile.
           "${GODOT}" --verbose --headless --disable-crash-handler --path project --export-debug "iOS Tests" "${{github.workspace}}/${{matrix.test-executable}}" || true
 
           if [ ! -f "${{github.workspace}}/${{matrix.test-executable}}" ]; then


limbonaut:
#461 was merged, which enabled running E2E on iOS in SauceLabs. Work on this one should be unblocked now.

exports/export_presets.cfg
```diff
@@ -241,17 +241,17 @@ script_export_mode=0
...
 architectures/arm64=true
-application/app_store_team_id="APP_STORE_TEAM_ID"
+application/app_store_team_id="97JCY7859U"
 application/export_method_debug=1
 application/code_sign_identity_debug=""
 application/code_sign_identity_release=""
-application/provisioning_profile_specifier_debug=""
-application/provisioning_profile_specifier_release=""
+application/provisioning_profile_specifier_debug="match Development io.sentry.SentryGodotProject"
+application/provisioning_profile_specifier_release="match Development io.sentry.SentryGodotProject"
```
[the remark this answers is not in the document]
limbonaut: We are not using that one -- we are signing code for Development and exporting as such for debug.

.github/workflows/test_integration.yml
```diff
@@ -51,14 +72,44 @@ jobs:
...
         run: |
-          "${GODOT}" --verbose --headless --disable-crash-handler --path project --install-android-build-template --export-debug "Android Tests" "${GITHUB_WORKSPACE}/exports/android.apk" || true
+          "${GODOT}" --verbose --headless --disable-crash-handler --path project --install-android-build-template --export-debug "Android Tests" "${{github.workspace}}/${{matrix.test-executable}}" || true
+
+      # Needed for fastlane
+      - name: "iOS: Setup Ruby"
+        if: matrix.export-platform == 'ios'
+        uses: ruby/setup-ruby@90be1154f987f4dc0fe0dd0feedac9e473aa4ba8 # v1.286.0
+        with:
+          bundler-cache: true
```
bitsandfoxes, line 89: I missed it, but why do we need fastlane?

.github/workflows/test_integration.yml
```diff
@@ -51,14 +72,44 @@ jobs:
...
+        run: bundle exec fastlane prepare_signing
+
+      - name: "iOS: Export project"
+        if: matrix.export-platform == 'ios'
+        timeout-minutes: 10
+        shell: bash
+        run: |
+          mkdir -p "$(dirname "${{github.workspace}}/${{matrix.test-executable}}")"
+
+          # TODO: Export fails because Godot attempts to build the Xcode project, which requires a certificate and provisioning profile.
```
bitsandfoxes: Is that TODO still relevant? For this to work there needs to be an output, right?

fastlane/Fastfile
bitsandfoxes, line 1: So this is all just for signing? 🙃

.github/workflows/test_integration.yml
bitsandfoxes: Signing. Gotcha.

fastlane/Fastfile
limbonaut: It's mostly code signing and CI, because part one was done in the previous PR here: https://github.com/getsentry/sentry-godot/pull/461 and when adding tests for Android. Same tests are reused on each platform with some small adjustments.

.github/workflows/test_integration.yml
limbonaut: Oh, need to remove that TODO, thanks!

Issue #114, opened 2025-02-12
Support for Web exports


bruno-garcia:
https://docs.godotengine.org/en/latest/tutorials/export/exporting_for_web.html

We don't need to bundle the JS SDK as part of this release.
For that wen can have a separate tracking issue listing the benefits of it.

Similar to:
* https://github.com/getsentry/sentry-unity/issues/668

oxeron:
Hello, is it already possible to export a web version of a godot project with the JS SDK manually ? Or should we have to wait for a specific godot version implementation ? I tried to add JS loader inside my web export template but I get this error: "[Warning] The Sentry loader you are trying to use isn't working anymore, check your configuration. (8e63051e6d04ec7fa26ef1d0f325f1c6.min.js, line 2)"

<img width="1294" alt="Image" src="https://github.com/user-attachments/assets/aafe600b-5c35-477f-9482-37808ed54b10" />

<img width="990" alt="Image" src="https://github.com/user-attachments/assets/0f841f02-76bc-4f77-bdcf-d7dffee1af9e" />

limbonaut:
@oxeron Sorry for late reply. I don't know if it's currently possible to export with a custom integration of JS SDK, but I've heard some folks trying something like that. We’ll be adding support for Web after 1.0.0 release, based on priorities. If you’d like this added, please give it a 👍!

oxeron:
Thanks for the information

StanleyAlbayeros:
@limbonaut web support is definitely something super useful! I'm doing a lot of game jams almost exclusively through web exports as those are more likely to get reviewed lately and it would be super helpful to get stats on that.

limbonaut:
@StanleyAlbayeros It's one of our priorities after 1.0.0 hits stable. If you’d like this added, please give it a 👍! It helps us gauge interest.

limbonaut:
Web support is mostly complete, the only remaining blockers for release are the docs and #530.

limbonaut:
Pre-release is out: https://github.com/getsentry/sentry-godot/releases

limbonaut:
Web support released with 1.4.0 🎉
Update to AssetLib is pending review.

PR #488, opened 2026-01-29
feat: Support Web platform

limbonaut:
This PR brings support for Web platform using Sentry JS SDK under the hood. This is a full-featured implementation supporting all our current functionality.

- Implements #114 

Sample issues:
- https://sentry-sdks.sentry.io/issues/6761298233/events/d89ac947de0042efa1934095d39b60a3/
- https://sentry-sdks.sentry.io/issues/6782593073/events/efb880c168c04db78ba0cfddd9a2cc66/
- https://limbonaut.sentry.io/issues/82592227/events/e10901e3c9b444a49a3e023d0d7a16bc/

This implementation uses bindings via JavaScriptBridge engine singleton and a small NPM TypeScript project that contains custom bridge code to ease interop between C++ and JS layers, and incorporates Sentry JavaScript SDK in the output: it bundles both the bridge and the JS SDK code in one minified file (sentry-bundle.js + map). So this bundle will be shipped with the libs and included in HTML template head, and also copied to the exported directory on project export (via editor plugin). All steps are automated, so no additional setup required from users, except for enabling "Extension Support" in the export options (I added a warning if it's not enabled). Also, I ended up sticking with WebPack+Terser plugin for tree shaking, as the build times were acceptable, however, the bundle size was a bit smaller than with esbuild.

I ran our tests via exported test project, and while some needed adjustments, all the relevant tests passed except the ones that need a fix in the JS SDK (fix [merged](https://github.com/getsentry/sentry-javascript/pull/19113) - waiting for release). We'll need to find a way to run these tests in CLI/CI, but currently, this can be done manually on a local computer by launching the test project in a Web browser and pressing corresponding UI buttons to run the tests. If you are trying to do this locally, you need to add Web export configuration first in "Export project..." dialog (enable "Extension support" there), and then you can press "Remote Deploy" icon in the top right corner, and then "Run in Browser".

This PR also support detecting native errors from WASM modules. However, WASM symbolication is failing even though symbols are uploaded and binaries contain valid debug IDs. The issue appears to be in the module detection, so debug images aren't showing up in `debug_meta.images`,and Sentry has nothing to match symbols against. But I'll need to investigate it further to confirm it. Our main bread and butter are runtime errors, so this shouldn't be a blocker.
- Sample WASM error: https://sentry-sdks.sentry.io/issues/7189020857/events/577170bd0f1c4edcb503c19ae959d6f1/
  - Note: Only symbols for libgamelogic are uploaded. Frames are recognized, but modules are missing.

Follow-ups:
- #519
- #517
- #496 
- #515
- #516
- #518

### DIFF
--- a/src/sentry/javascript/bridge/src/sentry-bridge.ts
+++ b/src/sentry/javascript/bridge/src/sentry-bridge.ts
@@ -75,7 +75,9 @@ class SentryBridge {
     enableLogs: boolean,
     sdkVersion: string,
   ): void {
-    console.log("Initializing Sentry via bridge...");
+    if (debug) {
+      console.log("Initializing Sentry via bridge...");
+    }
 
     const options: any = {
       dsn,


src/sentry/javascript/javascript_sdk.cpp
```diff
@@ -0,0 +1,322 @@
...
+		bytes.set(HEAPU8.subarray(data, data + size));
+
+		window.SentryBridge.addBytesAttachment(filenameStr, bytes, contentTypeStr);
+	} catch (e) {
+		console.error("Failed to add bytes attachment:", e);
+	}
+});
+
+// Emscripten JS function to push bytes directly to bridge layer and return id for later retrieval.
+EM_JS(uint8_t, add_bytes, (const uint8_t *data, int size), {
```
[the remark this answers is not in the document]
limbonaut: Fixed, good catch

src/sentry/level.cpp
```diff
@@ -21,6 +21,23 @@ godot::CharString level_as_cstring(Level level) {
...
+		case Level::LEVEL_WARNING:
+			return "warning";
+		case Level::LEVEL_ERROR:
+			return "error";
+		case Level::LEVEL_FATAL:
+			return "fatal";
+		default:
+			return "unknown";
+	}
+}
```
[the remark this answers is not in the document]
limbonaut: I'd rather avoid unnecessary memory ops.

limbonaut:
@sentry review

mujacica:
@cursor review

mujacica:
@sentry review

limbonaut:
I'll keep it open for some time, no problem @andreiborza. I expected that people might need some time to review it.
I'm going to do follow-ups in spin-off branches, so keeping it open is not blocking me.

src/editor/sentry_editor_export_plugin_web.cpp
```diff
@@ -0,0 +1,99 @@
...
+			if (!head_include.contains(line)) {
+				if (!adding) {
+					adding = true;
+					head_include += "\n" + html_head_comment;
+				}
+				head_include += "\n" + line;
+			}
+		}
+		get_export_preset()->set("html/head_include", head_include);
+	}
```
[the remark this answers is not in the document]
limbonaut: The one case we can't handle is the plugin being fully removed -- our code simply won't run. But that's a manual cleanup the user would be expected to do when uninstalling an addon. Our change to `head_include` is prominent in the export UI, with a helpful comment.

src/editor/sentry_editor_export_plugin_web.cpp
```diff
@@ -0,0 +1,99 @@
...
+		for (const String &line : html_head_content) {
+			if (!head_include.contains(line)) {
+				if (!adding) {
+					adding = true;
+					head_include += "\n" + html_head_comment;
+				}
+				head_include += "\n" + line;
+			}
+		}
+		get_export_preset()->set("html/head_include", head_include);
```
bitsandfoxes: I might be reading this wrong but doesn't this add to the preset with every run?

src/sentry/javascript/bridge/src/sentry-bridge.ts
```diff
@@ -0,0 +1,333 @@
...
+      debug,
+      release,
+      dist,
+      environment,
+      sampleRate,
+      maxBreadcrumbs,
+      enableLogs,
+      _metadata: {
+        sdk: {
+          name: "sentry.javascript.godot",
```
bitsandfoxes, line 93: 🎉

src/sentry/javascript/bridge/webpack.config.js
```diff
@@ -0,0 +1,48 @@
...
+	},
+	optimization : {
+		minimizer : [
+			new TerserPlugin({
+				terserOptions : {
+					format : {
+						comments : false
+					},
+					compress : {
+						drop_console : false,
```
bitsandfoxes, line 38: We're not dropping the console? Even in release?

src/sentry/sentry_sdk.cpp
```diff
@@ -368,7 +372,12 @@ void SentrySDK::prepare_and_auto_initialize() {
 	}
 #endif
 
-	_auto_initialize();
+	if (internal_sdk->get_capabilities().has_flag(InternalSDK::SUPPORTS_EARLY_INIT)) {
+		_auto_initialize();
+	} else {
+		// Defer automatic initialization when the underlying SDK cannot be initialized early.
+		callable_mp(this, &SentrySDK::_auto_initialize).call_deferred();
+	}
```
bitsandfoxes, line 380: So when running on the Web the SDK is initialized when the main loop starts. We should make sure to note this on the docs somewhere as this creates a "blindspot" between WASM module starting and the scene being loaded.

src/sentry/javascript/bridge/src/sentry-bridge.ts
```diff
@@ -0,0 +1,333 @@
...
+    dsn: string,
+    debug: boolean,
+    release: string,
+    dist: string,
+    environment: string,
+    sampleRate: number,
+    maxBreadcrumbs: number,
+    enableLogs: boolean,
+    sdkVersion: string,
+  ): void {
```
andreiborza, line 77: q: How about metrics? Could add `beforeSendMetric` and `enableMetrics`.

src/sentry/javascript/bridge/webpack.config.js
```diff
@@ -0,0 +1,48 @@
...
+						drop_debugger : true,
+						dead_code : true
+					}
+				},
+				extractComments : false
+			})
+		]
+	},
+	devtool : 'source-map'
```
andreiborza, line 47: q: How will the sourcemaps make it to Sentry? We have a bundler plugin for webpack that coud take care of that : https://www.npmjs.com/package/@sentry/webpack-plugin

src/sentry/javascript/bridge/src/sentry-bridge.ts
limbonaut: We didn't add metrics in Godot yet. Tracked here:
- #497 

I'd prefer to do it across all platforms in a separate PR.

src/sentry/javascript/bridge/webpack.config.js
limbonaut: In Godot SDK, we distribute a ready pre-compiled plugin. The builds are performed in CI, and both the bundle and sourcemap are distributed with the addon in a zip archive. Currently, there is no automation for uploading it to sentry.io, but I can see a possibility for an export plugin and an option to upload sourcemaps on Godot project export with the help of `sentry-cli` perhaps? I assume it knows how to do that.

src/editor/sentry_editor_export_plugin_web.cpp
limbonaut: It adds only if the expected lines are missing, but I see why this code may look confusing. I can rewrite it for clarity.

src/sentry/javascript/bridge/webpack.config.js
limbonaut: Hmm, maybe we should? In the bridge layer, we don't log much. But it's not just the bridge layer - JS SDK is also bundled here. Wdyt @andreiborza?

src/sentry/sentry_sdk.cpp
limbonaut: It still loads before the scene tree, I think. This is early initialization path (aka when you are not initializing from code). I'll need to test, but I'm pretty sure it's even before scripting goes up. Godot processes message queue regularly during setup.

src/sentry/javascript/javascript_sdk.cpp
```diff
@@ -0,0 +1,366 @@
...
+
+void JavaScriptSDK::close() {
+	ERR_FAIL_COND(js_sentry_bridge().is_null());
+
+	js_sentry_bridge()->call(JAVASCRIPT_SN(close));
+
+	_before_send_js_callback.unref();
+	_before_send_log_js_callback.unref();
+	file_attachments.clear();
+}
```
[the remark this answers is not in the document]
limbonaut: The beforeSend callback runs synchronously at capture time, not during close() flush.

src/sentry/javascript/bridge/webpack.config.js
andreiborza: I'd not drop them, all our console logs are guarded by the debug setting. You'd need them if users set `debug: true` in the init.

src/sentry/javascript/bridge/webpack.config.js
andreiborza: Yea, given that you'll have to upload more than just js sourcemaps, using sentry-cli directly is probably the way.

PR #490, opened 2026-02-03
ref: Extract hash funcs and decouple implementation from STL

limbonaut:
This PR replaces std::hash with FNV-1a hash implementation, and extracts hash functions into separate header file. This particular STL dependency was causing runtime errors with WASM compilation, because, since we compile as a side module, the main module needs to provide such compilation dependencies, and it proved to be problematic in my setup.
- Extracted from #488

### DIFF
--- a/src/sentry/util/hash.h
+++ b/src/sentry/util/hash.h
@@ -21,7 +21,7 @@ _FORCE_INLINE_ size_t fnv1a_hash(const char *p_data, size_t p_len) {
 	return hash;
 }
 
-_FORCE_INLINE_ size_t hash(std::string_view p_value) {
+_FORCE_INLINE_ size_t hash(const std::string_view &p_value) {
 	return fnv1a_hash(p_value.data(), p_value.size());
 }
 


src/sentry/util/hash.h
```diff
@@ -0,0 +1,38 @@
+#pragma once
+
+#include <godot_cpp/core/defs.hpp>
+#include <string_view>
+
+namespace sentry::util {
+
+// FNV-1a hash - portable hash implementation that doesn't rely on std::hash<*>.
```
jpnurmi, line 8: is it public domain?

limbonaut: You can't import our library directly, so nothing is really public unless explicitly registered within godot object system.

jpnurmi: I mean the license of the algorithm - seems so :)

> The FNV hash algorithms and reference FNV source code have been released into the public domain.

limbonaut: Oh, of course 🤦‍♂️  I misunderstood the question -  the algorithm is indeed in the public domain.

PR #492, opened 2026-02-05
ref: Use int64 for context values on native

limbonaut:
This PR changes some legacy code to use int64 for context values on native (and not only). Previously, we didn't have support for 64 bit values in sentry-native, so I used double as a workaround. This PR also adds checks to avoid adding -1 `device.memory` values, which results in processing errors in sentry.io (met this issue on Web platform).
- Extracted from #488

### DIFF
--- a/src/sentry/contexts.cpp
+++ b/src/sentry/contexts.cpp
@@ -137,7 +137,7 @@ Dictionary make_device_context(const Ref<RuntimeConfig> &p_runtime_config) {
 
 	auto dir = DirAccess::open("user://");
 	if (dir.is_valid()) {
-		device_context["free_storage"] = double(dir->get_space_left());
+		device_context["free_storage"] = dir->get_space_left();
 	}
 
 	device_context["processor_count"] = OS::get_singleton()->get_processor_count();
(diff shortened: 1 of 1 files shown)


src/sentry/contexts.cpp
```diff
@@ -148,24 +162,26 @@ Dictionary make_device_context_update() {
...
-#ifndef IOS_ENABLED
+#if !defined(IOS_ENABLED)
 	// NOTE: Memory info access on iOS can cause runtime errors in Godot 4.5.
 	const Dictionary &meminfo = OS::get_singleton()->get_memory_info();
-	// NOTE: Using double since int32 can't handle size in bytes.
-	//       On some platforms, memory info may not be available (-1).
-	int64_t free_memory = meminfo["free"];
-	if (free_memory >= 0) {
-		device_context["free_memory"] = double(free_memory);
+	int64_t mem_free = meminfo["free"];
```
jpnurmi, line 168: Not important, so feel free to ignore, but this looks very similar to the code above. Would a consolidated "memory context" helper make sense? Is this branch intentionally missing memory size?

jpnurmi (approved):

JoshuaMoelans (approved):

src/sentry/contexts.cpp
limbonaut: This method updates the device context values, that are volatile and can change from event to event, and since the physical memory size is not changing, it's not being updated.

limbonaut: There is a little bit of repetition here, but I wouldn't qualify it for "let's extract a function". Adding more abstraction just makes it harder to follow.

PR #500, opened 2026-02-10
ref(build): Produce iOS xcframework with framework bundle slices

limbonaut:
Produce iOS xcframeworks with proper `.framework` bundle slices instead of bare `.dylib` files.

Previously, `CreateXCFrameworkFromLibs` passed bare dylibs to `xcodebuild -create-xcframework -library`. Godot's export pipeline would then need to convert these into framework bundles via [`_convert_to_framework()`](https://github.com/godotengine/godot/blob/4.5.1-stable/editor/export/editor_export_platform_apple_embedded.cpp#L1042) during iOS export. By shipping proper framework slices from the start, we avoid this extra conversion step entirely.

Each dylib is now wrapped in a `.framework` bundle (with Info.plist and correct install name) before creating the xcframework with the `-framework` flag.

Also restores and improves the `plist` SCons tool (removed in cb8ae34) for generating framework Info.plist files.

### DIFF
--- a/site_scons/site_tools/plist.py
+++ b/site_scons/site_tools/plist.py
@@ -55,13 +55,13 @@ def _framework_plist_content(
 \t<key>CFBundlePackageType</key>
 \t<string>{bundle_package_type}</string>
 \t<key>CFBundleShortVersionString</key>
-\t<string>{bundle_version}</string>
+\t<string>{short_version}</string>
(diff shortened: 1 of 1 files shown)


limbonaut:
I've checked the artifact from CI - all good and it works fine on device.

PR #503, opened 2026-02-10
fix: Recreate SentryOptions on each init cycle

### LINKED ISSUE #501: SentryOptions retains stale values across close/init cycles
After switching from automatic-init-only to supporting explicit `init()`/`close()` cycles, `SentryOptions` retained its legacy singleton model. The singleton is created once at extension load and never recreated, so calling `SentrySDK.close()` followed by `SentrySDK.init()` causes old state to leak into the next session — callbacks, modified options, and event processors all persist.

Other Sentry SDKs create fresh options on each `init()` call.

A secondary issue: project settings values are loaded into the options in constructor, polluting default values visible to Godot's class reference XML doc generation.

limbonaut:
After switching from automatic-init-only to supporting explicit `init()`/`close()` cycles, `SentryOptions` retained its legacy singleton model. The singleton was created once at extension load and never recreated, so calling `SentrySDK.close()` followed by `SentrySDK.init()` caused old state to leak into the next session — callbacks, modified options, and event processors all persisted.

- Fixes #501

Move `SentryOptions` ownership from a static singleton to `SentrySDK`, so each `init()` call creates fresh options loaded from project settings. This prevents runtime values (`before_send`, event processors, etc.) from leaking across `close()`/`init()` cycles.

Key changes:
- Replace `SentryOptions` singleton with `create_from_project_settings()` factory
- `SentrySDK` owns options via `Ref<SentryOptions>` member, exposed through `get_options()` accessor and `SENTRY_OPTIONS()` macro
- `SentryExperimental` uses a raw owner pointer (with null-safety in destructor) instead of routing through a global accessor
- Built-in event processor registration moved into `init()` so processors don't accumulate on repeated cycles

### DIFF
--- a/.github/workflows/build_gdextension.yml
+++ b/.github/workflows/build_gdextension.yml
@@ -295,10 +295,11 @@ jobs:
         uses: actions/upload-artifact@v4
         with:
           name: sentry.${{matrix.platform}}.${{matrix.target}}.${{matrix.arch}}${{ contains(matrix.scons-flags, 'threads=no') && '.nothreads' || '' }}
+          # NOTE: Include-exclude pattern forces archive to start with project/ as root directory.
           path: |
             project/
-            !project/addons/gdUnit4
-            !project/test/
+            !project/
+            project/addons/sentry/
 
   android-plugin:
     name: 🤖 Android Plugin AARs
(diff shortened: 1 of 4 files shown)


mujacica (approved):
LGTM

PR #504, opened 2026-02-10
ci: Ship only addons/sentry/ in build artifacts and release zip

### LINKED ISSUE #502: Consider updating .gitattributes
Hey there! You may want to consider updating your .gitattributes file so users only download the asset itself; currently it downloads the full repo. A simple version would be something like:
```
# Normalize line endings for all files that Git considers text files.
* text=auto eol=lf

# Only include the addons folder when downloading from the Asset Library.
/**        export-ignore
/addons    !export-ignore
/addons/** !export-ignore
```

limbonaut:
Ship only `addons/sentry/` in CI build artifacts and the final release zip, instead of the full test project directory. This makes the release archive compatible with Godot AssetLib, which expects `addons/<plugin>/` at the root.

- Fixes #502

Previously, build artifacts included the entire `project/` directory (minus `test/` and `gdUnit4`), and the release zip contained everything in the artifact. Users installing from AssetLib would get demo scenes, test scripts, and project files alongside the actual addon. And this creates a lot of noise, and file conflicts in the existing projects.

### DIFF
--- a/.github/workflows/build_gdextension.yml
+++ b/.github/workflows/build_gdextension.yml
@@ -295,10 +295,11 @@ jobs:
         uses: actions/upload-artifact@v4
         with:
           name: sentry.${{matrix.platform}}.${{matrix.target}}.${{matrix.arch}}${{ contains(matrix.scons-flags, 'threads=no') && '.nothreads' || '' }}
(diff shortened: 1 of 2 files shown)


limbonaut:
> Nice, should we release/re-upload after this gets merged?

Yeah, let's merge the fixes and do a release.

PR #522, opened 2026-02-16
build: Add build.properties and split macOS CI into per-arch builds

limbonaut:
Add `build.properties` as a single source of truth for platform minimum versions and toolchain versions (`macos.deployment_target`, `ios.min_version`, `emscripten.version`). SConstruct reads these at build time with arch-specific override support, so both local builds and CI use the same values without duplication.

- Follow-up from #442 

Split macOS CI from a single `arch=universal` build into separate x86_64 and arm64 builds with a lipo step to combine them. This allows setting different `-mmacosx-version-min` per architecture — x86_64 targets macOS 10.14 (sentry-cocoa 9.0 minimum) while arm64 targets 11.0 (first Apple Silicon macOS).

Depends on the base branch (`chore/bump-cocoa-9.0.0`) being merged first — will rebase onto main after.

Changes:
- `build.properties` — new file with `macos.deployment_target`, `macos.deployment_target.x86_64`, `ios.min_version`, `emscripten.version`
- `site_scons/utils.py` — shared `read_property()` utility, replaces inline `get_property()` in sentry-cocoa.SConscript
- `SConstruct` — reads build.properties and sets SCons option defaults (arch-specific overrides, platform-guarded)
- CI workflow — removes hardcoded values, splits macOS into per-arch builds, adds `macos-universal` lipo job

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -89,8 +89,17 @@ def set_argument_default(scons_key, props_key):
                 pass
         ARGUMENTS[scons_key] = read_property(props_key, BUILD_PROPERTIES)
 
+# Detect platform (matching godot-cpp's auto-detection logic).
 platform = ARGUMENTS.get("platform", "")
-if platform == "macos" or (not platform and sys.platform == "darwin"):
+if not platform:
+    if sys.platform.startswith("linux"):
+        platform = "linux"
+    elif sys.platform == "darwin":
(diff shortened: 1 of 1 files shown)


limbonaut:
Looks like we can avoid using lipo. Closing.

PR #525, opened 2026-02-17
ci: Add JavaScript SDK dependency updater

### LINKED ISSUE #515: Update JavaScript SDK dependency automatically
JavaScript SDK dependency is tracked in the bridge package meta.

limbonaut:
Add automatic dependency tracking for the Sentry JavaScript SDK used in the Web platform bridge (`src/sentry/javascript/bridge/`).

Adds `scripts/javascript-version.ps1` — a version management script following the same interface as `android-version.ps1` (`get-version`, `get-repo`, `set-version`). It updates `@sentry/browser` and `@sentry/wasm` versions in `package.json` and runs `npm install` to keep the lockfile in sync.

Also adds a `sentry-javascript` job to the `update-deps` workflow using the new `ssh-key` input (see #524).

- Follow-up from #488
- Closes #515

### DIFF
--- a/scripts/javascript-version.ps1
+++ b/scripts/javascript-version.ps1
@@ -8,6 +8,12 @@ Set-StrictMode -Version latest
 $bridgeDir = "$PSScriptRoot/../src/sentry/javascript/bridge"
 $packageJsonFile = "$bridgeDir/package.json"
 
+# Sentry packages to track — add new packages here.
+$sentryPackages = @(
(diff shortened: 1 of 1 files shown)


limbonaut:
The test run has failed, but that's because the script is not in `main` and it created a merge conflict. Testing locally, it works as expected.

jpnurmi (approved):

PR #542, opened 2026-02-25
ci: Generate a separate package with demo project

### LINKED ISSUE #514: Generate a separate package with samples
We need to provide some sample code for users to try out different Sentry's functionality.

Previously, we were shipping our test project with the main package but it proved to be problematic with AssetLib distribution, so it was removed in #504.

limbonaut:
Generate a second zip in the packaging pipeline containing the demo project with compiled addon binaries, so users can try Sentry features out of the box.

- Closes #514
- Docs https://github.com/getsentry/sentry-docs/pull/16583

The demo project's `project.godot` is sanitized to remove gdUnit4 references (`[editor_plugins]` and `[gdunit4]` sections) and the DSN option. The `test/` directory is excluded.

Also renames the addon-only zip from `sentry-godot-gdextension-` to `sentry-godot-`.

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -90,10 +90,26 @@ jobs:
           # * Create release archive
           version=$(grep 'VERSION =' SConstruct | cut -d '"' -f 2)
           git_short_sha=$(git rev-parse --short HEAD)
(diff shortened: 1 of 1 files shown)


limbonaut:
We can add this package as separate entry in AssetLib for Tempate projects:
image

PR #553, opened 2026-03-02
feat: Metrics support

### LINKED ISSUE #497: Metrics Support
Add support for Metrics in Godot.

Depends on:
- #481 
- #487

limbonaut:
Adds support for emitting custom metrics to Sentry — track counters, gauges, and distributions directly from GDScript.

- Closes #497
- Docs PR getsentry/sentry-docs#16672

Metrics are enabled by default via `options.experimental.enable_metrics` and accessible through `SentrySDK.metrics`.

```gdscript
# Count occurrences
SentrySDK.metrics.count("match_started")
SentrySDK.metrics.count("enemies_spawned", wave_size, {"level": "castle"})

# Track point-in-time values
SentrySDK.metrics.gauge("players_online", lobby.get_player_count())
SentrySDK.metrics.gauge("memory_usage", OS.get_static_memory_usage(), "byte")

# Record value distributions
SentrySDK.metrics.distribution("level_load_time", load_time, "millisecond")
```

Metrics can be intercepted and modified via `options.experimental.before_send_metric`.

Supported on Windows, Linux, Android, and Web. Not yet supported on Apple platforms.

### DIFF
--- a/src/sentry/native/native_util.cpp
+++ b/src/sentry/native/native_util.cpp
@@ -163,13 +163,12 @@ sentry_value_t variant_to_attribute(const Variant &p_value) {
 }
 
 sentry_value_t dictionary_to_attributes(const Dictionary &p_attributes) {
-	if (p_attributes.is_empty()) {
-		return sentry_value_new_null();
-	}
 	sentry_value_t rv = sentry_value_new_object();
-	for (const Variant &key : p_attributes.keys()) {
-		sentry_value_set_by_key(rv, key.stringify().utf8(),
-				variant_to_attribute(p_attributes[key]));
+	if (!p_attributes.is_empty()) {
+		for (const Variant &key : p_attributes.keys()) {
+			sentry_value_set_by_key(rv, key.stringify().utf8(),
+					variant_to_attribute(p_attributes[key]));
+		}
 	}
 	return rv;
 }


android_lib/src/main/java/io/sentry/godotplugin/SentryAndroidGodotPlugin.kt
```diff
@@ -694,4 +769,91 @@ class SentryAndroidGodotPlugin(godot: Godot) : GodotPlugin(godot) {
...
+    @UsedByGodot
+    fun metricSetAttribute(handle: Int, attributeData: Dictionary) {
+        val metric = getMetric(handle) ?: return
+        val name = attributeData["name"] as String
+        val type = attributeData["type"] as String
+        val value = attributeData["value"]
+        val metricAttributes =
+            metric.attributes ?: HashMap<String, SentryLogEventAttributeValue>().also { metric.attributes = it }
+        metricAttributes[name] = SentryLogEventAttributeValue(type, value)
+    }
```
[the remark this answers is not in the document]
limbonaut: It's upstream decision to use that name.

src/sentry/native/native_util.cpp
```diff
@@ -162,4 +162,71 @@ sentry_value_t variant_to_attribute(const Variant &p_value) {
 	}
 }
 
+sentry_value_t dictionary_to_attributes(const Dictionary &p_attributes) {
+	sentry_value_t rv = sentry_value_new_object();
```
jpnurmi, line 166: `sentry_value_new_null` could avoid an allocation if the attributes dict is empty

src/sentry/internal_sdk.h
```diff
@@ -43,6 +43,10 @@ class InternalSDK {
 
 	virtual void add_attachment(const Ref<SentryAttachment> &p_attachment) = 0;
 
+	virtual void metrics_add_count(const String &p_name, int64_t p_value = 1, const Dictionary &p_attributes = Dictionary()) = 0;
```
jpnurmi: aren't the default values unnecessary noise in the InternalSDK API if the public wrapper already has them and always calls the internal SDK with all arguments?

src/sentry/native/native_util.cpp
limbonaut: That's how I initially implemented it, but the native was printing a message about it being null, so I made it an empty object instead.

src/sentry/internal_sdk.h
limbonaut: Makes sense, let me fix it.

src/sentry/processing/process_metric.cpp
```diff
@@ -0,0 +1,22 @@
...
+		return p_metric;
+	}
+
+	Ref<SentryMetric> processed = before_send_metric.call(p_metric);
+	if (processed == p_metric || processed.is_null()) {
+		return processed;
+	} else {
+		ERR_PRINT_ONCE("Sentry: before_send_metric callback must return the same metric object or null.");
+		return p_metric;
+	}
```
JoshuaMoelans: this is a nice check 👌 should we add the same to [logs](https://github.com/getsentry/sentry-godot/blob/main/src/sentry/processing/process_log.cpp#L13-L18)?

src/sentry/native/native_util.cpp
jpnurmi: Ah, indeed. Logs expect an object with the `logs_with_attributes` option turned on. Nevermind then :)

limbonaut: @jpnurmi https://github.com/getsentry/sentry-native/pull/1552

src/sentry/native/native_util.cpp
```diff
@@ -162,4 +162,72 @@ sentry_value_t variant_to_attribute(const Variant &p_value) {
...
+	if (p_attributes.is_empty()) {
+		return sentry_value_new_null();
+	}
+	sentry_value_t rv = sentry_value_new_object();
+	for (const Variant &key : p_attributes.keys()) {
+		sentry_value_set_by_key(rv, key.stringify().utf8(),
+				variant_to_attribute(p_attributes[key]));
+	}
+	return rv;
+}
```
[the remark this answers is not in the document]
limbonaut: Not really: https://github.com/getsentry/sentry-native/blob/923427f0b8dcb16cf0a6c4db2cbd70516e498e8e/include/sentry.h#L2203
Fix for the debug message is added in: https://github.com/getsentry/sentry-native/pull/1552

limbonaut:
I'll extract [865b84c](https://github.com/getsentry/sentry-godot/pull/553/commits/865b84c0f17a23aa779d39d005527d028c338ef2) as separate PR, so we don't have to wait on native release.

PR #562, opened 2026-03-05
feat: Add unit constants supported by Sentry specification

limbonaut:
This PR exposes `SentryUnit` class containing string constants for units supported in [specification](https://develop.sentry.dev/sdk/foundations/state-management/scopes/attributes/#units). They can be used with `gauge()` and `distribution()` like so:

```gdscript
SentrySDK.metrics.gauge("memory_usage", OS.get_static_memory_usage(), SentryUnit.byte)
SentrySDK.metrics.distribution("level_load_time", load_time, SentryUnit.millisecond)

# Custom units are also supported:
SentrySDK.metrics.gauge("items_in_inventory", item_count, "item")
```
- Follow-up from #553
- Docs https://github.com/getsentry/sentry-docs/pull/16794

### Units

| Category | Constants |
|----------|-----------|
| Duration | `nanosecond`, `microsecond`, `millisecond`, `second`, `minute`, `hour`, `day`, `week` |
| Information | `bit`, `byte`, `kilobyte`, `kibibyte`, `megabyte`, `mebibyte`, `gigabyte`, `gibibyte`, `terabyte`, `tebibyte`, `petabyte`, `pebibyte`, `exabyte`, `exbibyte` |
| Fraction | `ratio`, `percent` |

### Design clarification

The spec requires that SDKs SHOULD offer constants for supported units but MUST NOT restrict what can be sent (e.g. by using an enum).

Godot's binding API only supports integer constants (`BIND_CONSTANT` / `BIND_ENUM_CONSTANT`) — there is no mechanism to expose string constants to GDScript. To work around this, `SentryUnit` is registered as an engine singleton with read-only `String` properties, effectively acting as a static class with string constants. Each property name matches the wire format value directly (e.g., `SentryUnit.millisecond` returns `"millisecond"`). The metric methods keep their `String` parameter, so custom units work naturally.

### DIFF
--- a/project/test/isolated/test_metrics.gd
+++ b/project/test/isolated/test_metrics.gd
@@ -71,6 +71,20 @@ func test_distribution() -> void:
 	SentrySDK.metrics.distribution("response_time", 150.0)
 
 
+func test_gauge_with_sentry_unit() -> void:
+	metric_processed.connect(func(metric: SentryMetric):
+		assert_str(metric.unit).is_equal("byte")
+	, CONNECT_ONE_SHOT)
+	SentrySDK.metrics.gauge("memory_usage", 1024.0, SentryUnit.byte)
+
+
+func test_distribution_with_sentry_unit() -> void:
+	metric_processed.connect(func(metric: SentryMetric):
+		assert_str(metric.unit).is_equal("millisecond")
+	, CONNECT_ONE_SHOT)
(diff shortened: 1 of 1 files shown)


limbonaut:
Docs PR: https://github.com/getsentry/sentry-docs/pull/16794

jpnurmi (approved):

Issue #91, opened 2025-02-03
Support for C#


bruno-garcia:
Godot 4 includes a build with Mono built-in.
Sentry has support for C# in the [Sentry for .NET](https://github.com/getsentry/sentry-dotnet) as well as [Sentry for Unity](https://github.com/getsentry/sentry-unity) SDKs.

This ticket tracks support for C# on Godot games.

Skyppid:
Any status update on this? As someone loving Sentry and using Godot with C#, this would be highly appreciated :)

limbonaut:
@Skyppid The new estimate for the  initial .NET support is Q2-Q3 most likely. We are looking into how to best integrate .NET sdk into our internals.

cleardemon:
I noticed something with the recent release of Slay the Spire 2 (which is a Godot game, using C#). They seem to have included Sentry in their distribution, if you look at the macOS release ([depot exploration on SteamDB](https://steamdb.info/depot/2868842/)):

<img width="1916" height="320" alt="Image" src="https://github.com/user-attachments/assets/6c7afa2d-6767-4248-bd5a-91baafcbe2f1" />

I suspect they are just using the .NET version of Sentry, and may not be benefiting from the advantages that the Godot version would bring, but I thought it was an interesting observation.

Skyppid:
> I noticed something with the recent release of Slay the Spire 2 (which is a Godot game, using C#). They seem to have included Sentry in their distribution, if you look at the macOS release ([depot exploration on SteamDB](https://steamdb.info/depot/2868842/)):
[...]

PR #629, opened 2026-04-03
feat: Add C# support

limbonaut:
Adds .NET support to the SDK. Godot games written in C# can now use Sentry on par with GDScript projects, capturing exceptions, scopes, breadcrumbs, and everything else Sentry's .NET SDK supports. The .NET layer ships as a `Sentry.Godot` library inside the addon and can be wired into the user's `.csproj` via a single `<Import>` of `Sentry.Godot.props`.

- Refs #91

### How it works

The native C++ layer handles errors from GDScript and the engine, reports native crashes, and submits its own events. This PR adds a .NET layer that does the same for C# code, using Sentry's .NET SDK with its own transport, scopes, and event submission. The two layers are mostly independent - they connect for initialization, coordination and to share options at startup.

The native layer initializes during engine startup, before any .NET assembly loads. On auto-init, options are already resolved on the native side by the time .NET wakes up, so the .NET layer mirrors them across the native/managed boundary instead of re-resolving from scratch.

Initialization works from either side: whichever layer reaches `init` first signals the other, and a recursion guard makes the second call a no-op. Both auto-init (via `[ModuleInitializer]`) and manual `SentrySdk.Init(callback)` end up in the same state.

When Godot invokes managed user code, its managed bridge wraps the call in a try/catch and silently swallows any exception thrown. Godot logs an error, but the exception itself is lost. This PR adds first-chance exception handlers that detect those bridge catches and report them with `handled: false`.

Two handlers cover the cases: `LoggerExceptionHandler` is the default and works on all runtimes; `CoreClrExceptionHandler` is used when the Godot debugger is attached (which suppresses the logger path) and works on CoreCLR only.

On Godot API: The .NET layer doesn't depend on GodotSharp - calling into it early in the application lifecycle caused crashes on Android. Instead, the managed layer reaches Godot engine APIs through P/Invoke and native layer.

#### `Sentry.Godot.SentrySdk` does not expose other SentrySdk methods
Will be added in a separate PR. To address this, I want to try a code generation approach, which would make this PR even bigger than it has to be.

#### `Sentry.SentrySdk` is not internal
We add `Sentry.Godot.SentrySdk`, but this creates a footgun situation for users - the original `Sentry.SentrySdk` remains visible. Initializing Sentry through `Sentry.SentrySdk` will result in partial initialization. I plan to address this in a separate PR by adding a code analyzer which would produce a compilation warning if `Sentry.SentrySdk` is used directly.

### Follow-ups needed
- See #647


src/sentry/logging/sentry_godot_logger.cpp
```diff
@@ -289,6 +295,21 @@ void SentryGodotLogger::_log_error(const String &p_function, const String &p_fil
...
+	} else if (p_file.is_empty()) {
+		// File can be empty in MonoVM - use heuristic to detect C# errors.
+		for (int i = 0; i < p_script_backtraces.size(); i++) {
+			const Ref<ScriptBacktrace> &backtrace = p_script_backtraces[i];
+			if (backtrace->get_language_name() == "C#" && backtrace->get_frame_count() > 0) {
+				sentry::dotnet::handle_logger_error(p_file, p_code);
+				return;
+			}
+		}
+	}
```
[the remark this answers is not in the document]
limbonaut: Using same init order now, so native always initializes first.

project/addons/sentry/dotnet/Sentry.Godot/Interop/NativeBridge.cs
```diff
@@ -0,0 +1,341 @@
...
+            RepeatedErrorWindow = TimeSpan.FromMilliseconds(data.logger_limits.repeated_error_window_ms),
+            ThrottleEvents = data.logger_limits.throttle_events,
+            ThrottleWindow = TimeSpan.FromMilliseconds(data.logger_limits.throttle_window_ms),
+        };
+        opts.Dsn = data.dsn.TakeString() ?? "";
+        opts.Release = data.release.TakeString();
+        opts.Distribution = data.dist.TakeString();
+        opts.Environment = data.environment.TakeString();
+        opts.Debug = data.debug != 0;
+        opts.DiagnosticLevel = (SentryLevel)data.diagnostic_level;
```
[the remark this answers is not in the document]
limbonaut: Premise is wrong - SentryLevel is aligned and debug starts at 0.

project/addons/sentry/dotnet/Sentry.Godot/SentrySdk.cs
```diff
@@ -0,0 +1,139 @@
...
+
+    /// <summary>
+    /// Initializes the .NET SDK with an optional configuration callback.
+    /// </summary>
+    public static void Init(Action<SentryGodotOptions>? configureOptions = null)
+    {
+        if (_initializing)
+        {
+            return;
+        }
```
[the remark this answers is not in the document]
limbonaut: Init needs to be called on the main - this is non-issue.

project/addons/sentry/dotnet/Sentry.Godot/SentrySdk.cs
```diff
@@ -0,0 +1,139 @@
...
+            var godotOptions = new SentryGodotOptions();
+            godotOptions.ApplyNativeOptions();
+            godotOptions.ApplyTemplateSubstitutions();
+            InitDotnet(godotOptions);
+        }
+        finally
+        {
+            _initializing = false;
+        }
+    }
```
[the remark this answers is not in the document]
limbonaut: It will just reinit the hub, no big deal. Alternatives seem worse here.

limbonaut:
Testing two cases:
1. Exception handled in Godot bridge: Godot catches exceptions in try-catch, logs as errors, and the app continues running.
2. Unhandled exception from worker thread: the Godot game simply crashes.

**Linux**
- Bridge-handled: https://sentry-sdks.sentry.io/issues/7358364008/events/6008a967c3c343de8efc2029ff13ac99/
- Unhandled: https://sentry-sdks.sentry.io/issues/7358349731/events/d0e16593880446efb2248135d46a12c2/

**Android**
- Bridge-handled: https://sentry-sdks.sentry.io/issues/7398980973/events/276017f8f1104c8d8090d669998e5c46/
    - Note: The platform is detected as generic Linux, but it's actually Android.
- Unhandled: https://sentry-sdks.sentry.io/issues/7358349731/events/6e6e531fed4c4925bcb0be0bf4d54a5f/
    - Note: The app was supposed to crash, but it "hanged" instead. Probably crash handler issue?

**macOS**
- Bridge-handled: https://sentry-sdks.sentry.io/issues/7358364008/events/2497ab18c331481eb25396fbd07ce441/
- Unhandled: https://sentry-sdks.sentry.io/issues/7358349731/events/fabb5c3446974309865c4ee4d6ff9c0a/

**iOS**
- Bridge-handled: https://sentry-sdks.sentry.io/issues/7399300575/events/3c0e7ac6a7744f55b4d76c89a930761d/
- Unhandled: https://sentry-sdks.sentry.io/issues/7399309591/events/dd4a455ea48e43e988966cd4baa86a38/
  - Second event from Cocoa on app restart: https://sentry-sdks.sentry.io/issues/7399312202/events/04409a84c59d4381965818f0e5fb0153/ (I didn't upload symbols)

**Windows**
- Bridge-handled: https://sentry-sdks.sentry.io/issues/7358364008/events/1d7143bccbd6434ea26dd679c835543c/
- Unhandled: https://sentry-sdks.sentry.io/issues/7358349731/events/ee7fc80d11894785af29620a8c9db11a/
  - Second event from sentry-native: https://sentry-sdks.sentry.io/issues/7401196114/events/74cad0e5b513436267b6ec713880fd0d/

project/addons/sentry/dotnet/Sentry.Godot/ExceptionHandling/CoreClrExceptionHandler.cs
```diff
@@ -0,0 +1,169 @@
...
+        ("Godot.Bridge.CSharpInstanceBridge::", "Godot.Bridge.CSharpInstanceBridge"),
+        ("Godot.Bridge.ScriptManagerBridge::", "Godot.Bridge.ScriptManagerBridge"),
+        ("Godot.DelegateUtils::", "Godot.DelegateUtils"),
+        ("Godot.SignalAwaiter::", "Godot.SignalAwaiter"),
+        ("Godot.GD::", "Godot.GD"),
+    };
+
+    // Single lock guards both _threadExceptions and the per-thread queues.
+    private readonly object _lock = new();
+    private readonly Dictionary<long, Queue<Exception>> _threadExceptions = [];
```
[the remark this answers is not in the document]
limbonaut: Uncaught exceptions => crash, so it doesn't matter.

src/sentry/util/library_path.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#include <godot_cpp/variant/string.hpp>
+
+namespace sentry::util {
+
+inline godot::String get_gdextension_library_path() {
+	godot::String library_path;
+	godot::internal::gdextension_interface_get_library_path(godot::internal::library, library_path._native_ptr());
```
jpnurmi, line 10: where is this declared? transitive include via `string.hpp`? could we include the appropriate header?

src/sentry/logging/sentry_godot_logger.cpp
```diff
@@ -44,7 +45,12 @@ class RecursionGuard {
 };
 
 bool _get_script_context(const String &p_file, int p_line, String &r_context_line, PackedStringArray &r_pre_context, PackedStringArray &r_post_context) {
-	if (p_file.is_empty()) {
+	// Don't fetch C# context - fails to load and leads to errors.
+	if (p_file.is_empty() || p_file.to_lower().ends_with(".cs")) {
```
jpnurmi: nit-pick: transforms/allocates the entire path to lowercase to check the suffix. not as convenient, but maybe rfind(".") or only checking the last chars would be better?

src/sentry/logging/sentry_godot_logger.cpp
```diff
@@ -289,6 +295,21 @@ void SentryGodotLogger::_log_error(const String &p_function, const String &p_fil
 		return;
 	}
 
+	// Forward C# exceptions to the .NET layer for capture.
+	if (p_file.to_lower().ends_with(".cs")) {
```
jpnurmi: ditto

src/sentry/dotnet/csharp_interop.cpp
```diff
@@ -0,0 +1,295 @@
...
+}
+
+struct LoggerLimitsData {
+	int32_t events_per_frame;
+	int32_t repeated_error_window_ms;
+	int32_t throttle_events;
+	int32_t throttle_window_ms;
+};
+
+struct NativeOptions {
```
jpnurmi, line 51: how about some comment/reference to NativeBridge.cs?

project/addons/sentry/dotnet/Sentry.Godot/Interop/NativeBridge.cs
```diff
@@ -0,0 +1,347 @@
...
+    private struct LoggerLimitsData
+    {
+        public int events_per_frame;
+        public int repeated_error_window_ms;
+        public int throttle_events;
+        public int throttle_window_ms;
+    }
+
+    [StructLayout(LayoutKind.Sequential)]
+    private struct NativeOptions
```
jpnurmi, line 57: how about some comment/reference to csharp_interop.cpp?

project/addons/sentry/dotnet/Sentry.Godot.props
```diff
@@ -0,0 +1,15 @@
+<!--
+  Sentry SDK for Godot - .NET integration
+
+  Add the following line to your .csproj to enable:
+    <Import Project="addons/sentry/dotnet/Sentry.Godot.props" />
+
+  TODO: Auto-include this file in the user's project.
```
jpnurmi, line 7: i'm not an msbuild expert (nor a fan :P) but sentry-dotnet uses "buildTransitive" files for similar things - would something like that work here or are they only for nuget packages? 🤔 

- https://github.com/getsentry/sentry-dotnet/blob/main/src/Sentry/buildTransitive/Sentry.props
- https://github.com/getsentry/sentry-dotnet/blob/main/src/Sentry/Platforms/Native/buildTransitive/Sentry.Native.targets

project/addons/sentry/dotnet/Sentry.Godot/Sentry.Godot.csproj
```diff
@@ -0,0 +1,11 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFrameworks>net8.0;net9.0</TargetFrameworks>
```
jpnurmi: is net10.0 supported?

jpnurmi (commented):
I hope the .NET experts will also chime in. I focused mostly on the C++ side of things :)

src/sentry/util/library_path.h
limbonaut: Yes, it's a transitive include declared in `godot.hpp` pulled by anything from godotcpp. Let's make it explicit. FYI, this is internal namespace so there is a risk of breakage, but alternatives are not great either.

src/sentry/logging/sentry_godot_logger.cpp
limbonaut: Added optimized `ends_with_nocase_ascii()` for such needs.

src/sentry/dotnet/csharp_interop.cpp
limbonaut: Added respective comments in both files.

project/addons/sentry/dotnet/Sentry.Godot/Interop/NativeBridge.cs
limbonaut: Added respective comments in both files.

project/addons/sentry/dotnet/Sentry.Godot/Sentry.Godot.csproj
limbonaut: Added it.

project/addons/sentry/dotnet/Sentry.Godot/Interop/NativeBridge.cs
```diff
@@ -0,0 +1,351 @@
...
+                    return null;
+                }
+                return Encoding.UTF32.GetString((byte*)Ptr, (int)Len * SizeOfChar32);
+            }
+            finally
+            {
+                csharp_interop_string_free(Handle);
+            }
+        }
+    }
```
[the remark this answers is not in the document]
limbonaut: Actually, no. Examining the code here:
https://github.com/dotnet/runtime/blob/422c41a676096d58d1218bc4385289338cdac615/src/coreclr/vm/classlayoutinfo.cpp#L67
The alignment should match C ABI.

project/addons/sentry/dotnet/Sentry.Godot/Interop/NativeBridge.cs
```diff
@@ -0,0 +1,351 @@
...
+        public byte app_hang_tracking;
+        public double app_hang_timeout_sec;
+        public byte logger_enabled;
+        public byte logger_include_source;
+        public byte logger_include_variables;
+        public byte logger_messages_as_breadcrumbs;
+        public int logger_event_mask;
+        public int logger_breadcrumb_mask;
+        public byte enable_metrics;
+    }
```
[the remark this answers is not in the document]
limbonaut: This is correct: C# matches C ABI for field alignment, which I verified in source files.

project/addons/sentry/dotnet/Sentry.Godot/ExceptionHandling/CoreClrExceptionHandler.cs
```diff
@@ -0,0 +1,169 @@
...
+                if (_threadExceptions.TryGetValue(sourceOsTid, out var queue))
+                {
+                    dequeued = queue.Dequeue();
+                    remainingInQueue = queue.Count;
+                    if (remainingInQueue == 0)
+                    {
+                        _threadExceptions.Remove(sourceOsTid);
+                    }
+                }
+            }
```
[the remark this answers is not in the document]
limbonaut: This looks legit - need to fix this.

project/addons/sentry/dotnet/Sentry.Godot/Interop/NativeBridge.cs
```diff
@@ -0,0 +1,351 @@
...
+
+    private static Action<string, string>? _loggerErrorHandler;
+
+    [UnmanagedCallersOnly(CallConvs = new[] { typeof(System.Runtime.CompilerServices.CallConvCdecl) })]
+    private static unsafe void LoggerErrorCallback(char* code, int codeLen, char* file, int fileLen)
+    {
+        _loggerErrorHandler?.Invoke(
+                new string(file, 0, fileLen),
+                new string(code, 0, codeLen));
+    }
```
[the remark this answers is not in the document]
limbonaut: https://learn.microsoft.com/en-us/dotnet/standard/native-interop/exceptions-interoperability seems to confirm this. Just to be safe, I'll wrap those into try-catch.

project/addons/sentry/dotnet/Sentry.Godot/SentrySdk.cs
```diff
@@ -0,0 +1,139 @@
...
+    /// Always called by the native SDK after initialization, but should only take
+    /// effect when the native layer is auto-initialized or initialized from GDScript.
+    /// </remarks>
+    internal static void InitFromNative()
+    {
+        if (_initializing)
+        {
+            return;
+        }
+        _initializing = true;
```
jamescrosswell, line 68: I think there's a race condition here no?

I see from your [reply above](https://github.com/getsentry/sentry-godot/pull/629/changes#r3057739798) to a similar review comment that init needs to happen on main so it's a non issue... in which case can the check to see if `_initializing` is already happening be skipped entirely (it could never happen right). 

So I'm thinking:
- Either a race is possible, in which case the code needs changing
- Or a race isn't possible, in which case the code is redundant and can be removed

project/addons/sentry/dotnet/Sentry.Godot/Internal/SentryGodotInitializer.cs
```diff
@@ -0,0 +1,20 @@
...
+/// </summary>
+public static class SentryGodotInitializer
+{
+    public static void AutoInit()
+    {
+        NativeBridge.RegisterDotnetInit();
+        if (NativeBridge.IsEnabled())
+        {
+            GodotLog.Debug("Native is already initialized, proceeding with .NET initialization.");
+            SentrySdk.InitFromNative();
```
jamescrosswell, line 25: There's no chance the return value of `NativeBridge.IsEnabled` could change between line 14 and line 17 executing? E.g. the native initialisation completes after line 14 executes but before line 17 executes?

If the native initialisation does always happen in the same thread that this method executes, I'm wondering why the order of execution isn't deterministic (i.e. one should always fire before the other in that case right - in which case you could remove the alternative code path for initialisation).

project/addons/sentry/dotnet/Sentry.Godot/ExceptionHandling/LoggerExceptionHandler.cs
```diff
@@ -0,0 +1,85 @@
...
+    private static void OnLogError(string file, string code)
+    {
+        if (_isProcessing || _lastException == null)
+        {
+            return;
+        }
+
+        // Validate that _LogError was triggered by _lastException, not a stale one.
+        // The bridge passes the exception type name (and possibly message) as the "code" parameter.
+        var expectedType = _lastException.GetType().FullName;
```
jamescrosswell, line 54: Another race condition question... isn't it possible that `OnFirstChanceException` fires multiple times in a multithreaded environment, and `_lastException` is simply stored as a `static`... so the value of `_lastException` could change between line 47 and here. 

It might be better to capture the value of `_lastException` in a local variable before checking to see whether it's null or not (and before using it here)... pretty much exactly what's happening on line 65 - but pull that forward.

jamescrosswell (commented):
I didn't do an exhaustive review... partly since I'm not familiar enough with Godot to know if the comments I've made so far are useful or not. Probably I've asked questions more than provided feedback 😁

project/addons/sentry/dotnet/Sentry.Godot/SentrySdk.cs
limbonaut: This is supposed to be a re-entry guard, because initialization request can come from both ends: either managed or native side.

I made a chart to explain the expected initialization scenarios. Typical games are created either with GDScript or C#, and both can be mixed in the same project. We expect the initialization request to come from any of those scripting languages. I bundle both GDScript and C++ as "Native layer" here.

Screenshot 2026-04-14 at 13 59 11

I used dashed lines for steps that may never happen, because Godot comes in two flavors: standard and .NET edition. So, the .NET steps may not be used at all, either because we are running standard edition or when the engine doesn't load managed assemblies (no C# scripts loaded in the project).

project/addons/sentry/dotnet/Sentry.Godot/Internal/SentryGodotInitializer.cs
limbonaut: Native initialization is expected to happen on the same thread, as can be seen on the [initialization chart](https://github.com/getsentry/sentry-godot/pull/629#discussion_r3079253841). Even during automatic initialization, as I verified manually on different platforms, the `[ModuleInitializer]` is invoked on the main thread.
(Should probably add a test for ModuleInitializer, if possible, when I get to testing)

.NET assembly load doesn't have a guaranteed lifecycle, and in my tests I found various cases where assembly loaded much later or earlier than I expected. So this check verifies basically if we are already initialized in the native layer, and if so, we proceed to initialize the managed layer. This is expected to happen either during automatic initialization or when the assembly is loaded after the GDScript code has already initialized the SDK.

project/addons/sentry/dotnet/Sentry.Godot/ExceptionHandling/LoggerExceptionHandler.cs
limbonaut: `_lastException` is marked with `[ThreadStatic]` attribute, which means that each thread should get its own separate storage field. Both FirstChanceException and Godot's logger run on the same thread for the same exception, so there should be no race.

project/addons/sentry/dotnet/Sentry.Godot.props
limbonaut: From what I gather, buildTransitive is only used if you build as NuGet package, but the plan is to ship source code instead.

I think, we can auto-create a `Directory.Build.props` next to the user's .csproj. Just need some plan if the user already has such file - we could parse XML and inject our instructions if it exists.

project/addons/sentry/dotnet/Sentry.Godot/SentrySdk.cs
jamescrosswell: OK so is there a chance that both native and c# could be trying to initialise at the same time? If there is, then I think there's a race condition here. If there isn't, then this code could be removed entirely (since `_initializing` would always be false when entering this method). 

If we're not sure, maybe best to err on the side of caution and assume concurrency is possible, then add a double lock or something to prevent a race condition:
```
    internal static void InitFromNative()
    {
        if (_initializing)
        {
            return;
        }
        lock (_initLock)
        {
            if (_initializing)
            {
                return;
            }
            _initializing = true;
            try
            {
                var godotOptions = new SentryGodotOptions();
                godotOptions.ApplyNativeOptions();
                godotOptions.ApplyTemplateSubstitutions();
                InitDotnet(godotOptions);
            }
            finally
            {
                _initializing = false;
            }
        }
    }
```

Since the code only runs once on startup and contention is extremely unlikely, any overhead associated with the lock will be trivial.

project/addons/sentry/dotnet/Sentry.Godot/ExceptionHandling/LoggerExceptionHandler.cs
jamescrosswell: My bad - didn't see the `[ThreadStatic]` 👍🏻

project/addons/sentry/dotnet/Sentry.Godot/SentrySdk.cs
limbonaut: Let's trace how `InitFromNative` is actually used (for Scenario C on the chart):
1. `Init()` is called on the .NET side: 
https://github.com/getsentry/sentry-godot/blob/0335a2442c363336d357b6cbc6d42ffe7077f1b1/project/addons/sentry/dotnet/Sentry.Godot/SentrySdk.cs#L19
2. Which triggers native initialization here:
https://github.com/getsentry/sentry-godot/blob/0335a2442c363336d357b6cbc6d42ffe7077f1b1/project/addons/sentry/dotnet/Sentry.Godot/SentrySdk.cs#L39-L40
https://github.com/getsentry/sentry-godot/blob/0335a2442c363336d357b6cbc6d42ffe7077f1b1/project/addons/sentry/dotnet/Sentry.Godot/SentrySdk.cs#L87-L94
3. This calls `csharp_interop_sdk_init` on the native side, and triggers native `init()`:
https://github.com/getsentry/sentry-godot/blob/0335a2442c363336d357b6cbc6d42ffe7077f1b1/src/sentry/dotnet/csharp_interop.cpp#L257-L261
4. The native `init()` is designed to signal the .NET layer to initialize, given that it can also be called from GDScript (see Scenario B):
https://github.com/getsentry/sentry-godot/blob/0335a2442c363336d357b6cbc6d42ffe7077f1b1/src/sentry/sentry_sdk.cpp#L186-L187
https://github.com/getsentry/sentry-godot/blob/0335a2442c363336d357b6cbc6d42ffe7077f1b1/project/addons/sentry/dotnet/Sentry.Godot/Interop/NativeBridge.cs#L176-L187
5. Which calls `InitDotnet` competing with initial `Init` call (line 40):
https://github.com/getsentry/sentry-godot/blob/0335a2442c363336d357b6cbc6d42ffe7077f1b1/project/addons/sentry/dotnet/Sentry.Godot/SentrySdk.cs#L67

This call chain is why we need the re-entry guard. Since everything happens within the same thread, locking probably isn’t necessary, but I should probably add a comment to make intention clear.

project/addons/sentry/dotnet/Sentry.Godot/ExceptionHandling/CoreClrExceptionHandler.cs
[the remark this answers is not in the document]
limbonaut: Fixed in ab3c7e75c02fd3a4c0b2b5dcc28b01fac85912da

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Features
+
+- Add C# (.NET) support ([#629](https://github.com/getsentry/sentry-godot/pull/629))
```
jpnurmi: Is it still experimental?

jpnurmi (approved):
Looks nice and well-structured. I lost track if there are any unresolved findings besides the agreed follow-ups. IMO this looks like a great start 🎉

CHANGELOG.md
limbonaut: It really is. I don’t think we should release it until we’ve addressed some of the necessary follow-ups. I’ve started a changelog entry, but C# support is not complete. I think we’ll just keep adding PR links to that record as we merge more changes. And if it’s still experimental by the time we release, we’ll mark it accordingly.

jpnurmi: sounds good!

project/addons/sentry/dotnet/Sentry.Godot/ExceptionHandling/CoreClrExceptionHandler.cs
```diff
@@ -0,0 +1,183 @@
...
+        {
+            HandleExceptionCatchStart(eventData);
+        }
+    }
+
+    private void HandleExceptionCatchStart(EventWrittenEventArgs eventData)
+    {
+        var sourceOsTid = eventData.OSThreadId;
+        var catchTimestamp = eventData.TimeStamp;
+        string? bridgeName = eventData.Payload?[2] is string methodName ? GetGodotBridgeName(methodName) : null;
```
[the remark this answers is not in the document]
limbonaut: Event payload is documented and stable - this should not be an issue.

limbonaut:
I found issues with `CoreClrExceptionHandler` which I'll address in a separate PR, so it's easier to review.
- https://github.com/getsentry/sentry-godot/pull/644

project/addons/sentry/dotnet/Sentry.Godot/Interop/NativeBridge.cs
```diff
@@ -0,0 +1,365 @@
...
+            ThrottleEvents = data.logger_limits.throttle_events,
+            ThrottleWindow = TimeSpan.FromMilliseconds(data.logger_limits.throttle_window_ms),
+        };
+        opts.Dsn = data.dsn.TakeString() ?? "";
+        opts.Release = data.release.TakeString();
+        opts.Distribution = data.dist.TakeString();
+        opts.Environment = data.environment.TakeString();
+        opts.Debug = data.debug != 0;
+        opts.DiagnosticLevel = (SentryLevel)data.diagnostic_level;
+        opts.SampleRate = (float)data.sample_rate;
```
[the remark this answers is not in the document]
limbonaut: SameRate is float in .NET

Issue #648, opened 2026-04-21
Generate `SentrySdk.Dotnet.cs` to delegate upstream `Sentry.SentrySdk` public methods


limbonaut:
**\[GA blocker\]**

`Sentry.Godot.SentrySdk` currently only exposes `Init`. All other public entry points on upstream `Sentry.SentrySdk` (e.g. `CaptureException`, `CaptureMessage`, `ConfigureScope`, `AddBreadcrumb`, `StartTransaction`, `Flush`, …) are not reachable through the Godot wrapper, so users have to reach into `Sentry.SentrySdk` directly — which interacts badly with the initialization model (see [GODOT-188](https://linear.app/getsentry/issue/GODOT-188/add-a-code-analyzer-to-discourage-direct-sentrysentrysdk-usage)).

We can try code generation approach that enumerates `Sentry.SentrySdk` public members and emits a thin delegating partial class `SentrySdk.Dotnet.cs`. Doing it by hand would drift with upstream Sentry .NET releases.

limbonaut:
Completed in #657

PR #658, opened 2026-04-22
test(dotnet): Add snapshot test for SentrySdk source generator

limbonaut:
Add a snapshot test that pins the SentrySdk source generator's output, plus CI to run it on every PR.

- Refs #648
- Depends on #657

The generator emits a forwarder for every public static member of `Sentry.SentrySdk` — without coverage, drift from either side (a generator change or an upstream package update) would land silently. The test compiles a synthetic input, drives the generator, and diffs the produced source against a committed `.verified.txt` baseline. Intentional changes are accepted by renaming `*.received.txt` over `*.verified.txt`.

CI also compiles the demo project alongside the snapshot test. The snapshot catches emission drift, and the demo compile catches the downstream consequence — generator output that parses standalone but breaks when consumed (unresolved types, ambiguous overloads). On snapshot failure the workflow uploads `*.received.txt` as an artifact so reviewers can pull the diff without checking out the branch.

Failure output uses `Verify.DiffPlex` to show an inline unified diff in the test log rather than dumping the full received file, keeping CI logs readable.

Test project lives at `tests/dotnet/Sentry.Godot.SourceGenerators.Tests/` rather than under the addon directory, so dev-only files stay out of what ships to Godot users. Structure follows sentry-dotnet conventions: xUnit + Verify, flat `.verified.txt` filenames, programmatic pre-asserts.


limbonaut:
Check out an intentionally failed run: https://github.com/getsentry/sentry-godot/actions/runs/24792560778?pr=658

.github/workflows/test_dotnet.yml
```diff
@@ -0,0 +1,94 @@
...
+        with:
+          submodules: false
+
+      - name: Set up .NET
+        uses: actions/setup-dotnet@c2fa09f4bde5ebb9d1777cf28262a3eb3db3ced7 # v5.2.0
+        with:
+          dotnet-version: 8.0.x
+
+      - name: Build demo project
+        run: dotnet build "project/Sentry demo project.csproj" --nologo
```
jpnurmi: nit: already has `DOTNET_NOLOGO=true` in the global env-block

```suggestion
        run: dotnet build "project/Sentry demo project.csproj"
```

jpnurmi (approved):
Sweet!

mujacica (approved):

PR #671, opened 2026-04-29
feat(dotnet): Import `Sentry.Godot.props` into user's C# project automatically

### LINKED ISSUE #653: Auto-import `Sentry.Godot.props` into user's `.csproj`
**[GA blocker]**

Today the user has to manually add `<Import Project="addons/sentry/dotnet/Sentry.Godot.props" />` to their `.csproj`. That's a friction point and an easy step to miss — silently producing a non-working SDK.

Consider using `Directory.Build.props` for this.

limbonaut:
When a .NET project is detected, the editor now patches the user's C# project file (`.csproj`) to import the SDK's `Sentry.Godot.props` automatically, so the .NET build picks up Sentry's library without requiring a manual edit. 

- Closes #653
- Tests in #675 

The patcher uses a small XML-aware scanner (just enough to find `</Project>` and existing `<Import>` siblings), inserts a conditional `<Import Project="..." Condition="Exists(...)">` before `</Project>` if not already present, and detects the file's indentation and line ending so the inserted line matches the surrounding style, backs up the original csproj, and uses an atomic temp-file rename so the file is never left half-written.

### DIFF
--- a/src/editor/csproj_patcher.cpp
+++ b/src/editor/csproj_patcher.cpp
@@ -131,8 +131,23 @@ std::string_view _read_attribute_value(std::string_view p_tag, std::string_view 
 }
 
 bool _is_same_path(std::string_view p_path1, std::string_view p_path2) {
-	// TODO: Must handle different path separators.
-	return p_path1 == p_path2;
+	if (p_path1.size() != p_path2.size()) {
+		return false;
+	}
+	for (size_t i = 0; i < p_path1.size(); ++i) {
+		const char c1 = p_path1[i];
+		const char c2 = p_path2[i];
+		if (c1 == c2) {
+			continue;
+		}
+		const bool c1_is_sep = (c1 == '/' || c1 == '\\');
+		const bool c2_is_sep = (c2 == '/' || c2 == '\\');
+		if (c1_is_sep && c2_is_sep) {
+			continue;
+		}
+		return false;
+	}
+	return true;
 }
 
 inline void _vector_append(std::vector<uint8_t> &p_vec, std::string_view p_str) {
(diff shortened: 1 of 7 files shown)


mujacica (approved):

limbonaut:
Tests are added in: 
- #675

jpnurmi (commented):
You probably already looked into this, but is there maybe some .NET or MSBuild tooling or API that could give us a way out of needing to maintain a hand-rolled XML parser? 🤔

src/editor/csproj_patcher.cpp
jpnurmi, line 1: Are `src/editor/*` only built in editor mode? I don't see any build system changes, so it probably picks these with wildcards?

limbonaut: `src/editor/*.cpp` is globbed for `target=editor` builds

limbonaut: But just in case, I've added `TOOLS_ENABLED` guards.

jpnurmi (approved):
> You probably already looked into this, but is there maybe some .NET or MSBuild tooling or API that could give us a way out of needing to maintain a hand-rolled XML parser? 🤔

Or perhaps I'm thinking about this in a wrong way. Even if we're targeting .NET, we're not really fully in the .NET land. If we need to do this from C++, spinning off a managed runtime as a subprocess just to insert a one-liner might not be any better... 😭

limbonaut:
@jpnurmi
> Or perhaps I'm thinking about this in a wrong way. Even if we're targeting .NET, we're not really fully in the .NET land. If we need to do this from C++, spinning off a managed runtime as a subprocess just to insert a one-liner might not be any better... 😭

Yeah, I don't think we can access Godot's hostfxr context. I was also considering a small project that we could pre-compile as CLI tool and ship with the addon. We would run it from GDExtension lib with popen ala `sentry-cli`. But that would shift complexity into build system and packaging, increase our archive in size and start depending on user's dotnet setup. The 300 LoC scanner seems like a lesser evil to me, and to make it more robust, I'm adding CPP tests.

PR #675, opened 2026-04-30
test(cpp): Add C++ test runner using `doctest` and `CsprojPatcher` tests

### LINKED ISSUE #674: Add C++ unit tests
Sometimes, we want to test C++ code that is not reachable from public API, such as #671.
Let's add unit tests specifically for C++ that can take internal code and test it.

limbonaut:
This PR:
- Adds [doctest](https://github.com/doctest/doctest) as submodule (single-header library).
- Adds a test runner that can execute tests inside the engine (GDExtension lib can't run independent).
- Adds a test suite for `CsprojPatcher` added in #671 
- Adds CI workflow to run these tests.
- Closes #674 

Currently, we are testing our functionality using a GDScript testing framework. While that approach helps verify our public interfaces the same way they would be used by developers, it does limit us to GDScript-exposed interface. The C++ testing will help cover our internals that are not accessible from public API.


limbonaut:
@sentry review

mujacica (approved):

jpnurmi (approved):
Is a differently formatted `<Import>` worth testing? 🤔 
```
<Import Project="Sentry.Godot.props"
        Condition="Exists('Sentry.Godot.props')" />
```

Also, technically `<Import>` could be moved inside an `<ImportGroup>`, but I'm not sure if it's worth supporting or would even affect the parser or replacement in any way.

PR #678, opened 2026-05-04
test: Fix time tolerance issue in breadcrumb test on Windows

limbonaut:
This PR increases the tolerance in `test_breadcrumb_timestamp_is_set_automatically` from 1 ms to 50 ms to fix flaky failures on Windows. Godot's `Time.get_unix_time_from_system()` is backed by `GetSystemTime`, which has coarse resolution, while sentry-native captures breadcrumb timestamps via `GetSystemTimePreciseAsFileTime`, so the two readings can disagree by up to a system-timer tick.

### DIFF
--- a/project/test/suites/test_breadcrumb.gd
+++ b/project/test/suites/test_breadcrumb.gd
@@ -49,7 +49,9 @@ func test_breadcrumb_default_values() -> void:
 
 
 func test_breadcrumb_timestamp_is_set_automatically() -> void:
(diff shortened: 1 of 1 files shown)


limbonaut:
Failed run with this issue: https://github.com/getsentry/sentry-godot/actions/runs/25321027753/job/74232515706?pr=677

jpnurmi (approved):
Was it me who insisted on the shortest possible tolerance? 🙈

PR #696, opened 2026-05-08
feat(dotnet): Sync trace context with .NET layer

### LINKED ISSUE #652: Share trace ID between native and .NET layers
**[GA blocker]**

Events captured by the native and managed layers in the same runtime session are currently unrelated in Sentry. This matters even for C#-only projects: engine errors, native crashes, and anything captured by the native logger go out through the native layer, so without a shared trace context they can't be correlated with C# exceptions from the same user action.

We should pick one side to own the current trace context (likely native) and propagate `trace_id` to the other — wiring this through the same native/managed interop. An e2e test that produces correlated events from both layers and asserts a matching trace id would prove it works.

limbonaut:
Managed and native events need to be on the same trace. Otherwise, managed events don't see native errors and logs, and vice versa.
- Closes #652

### DIFF
--- a/src/sentry/cocoa/cocoa_sdk.mm
+++ b/src/sentry/cocoa/cocoa_sdk.mm
@@ -287,6 +287,17 @@ void CocoaSDK::remove_attribute(const String &p_name) {
 	}];
 }
 
+void CocoaSDK::set_trace(const String &p_trace_id, const String &p_parent_span_id) {
+	ERR_FAIL_COND(p_trace_id.is_empty());
+
+	objc::SentryId *trace_id = [[objc::SentryId alloc] initWithUUIDString:string_to_objc(p_trace_id)];
+	objc::SentrySpanId *span_id = p_parent_span_id.is_empty()
+			? [[objc::SentrySpanId alloc] init]
+			: [[objc::SentrySpanId alloc] initWithValue:string_to_objc(p_parent_span_id)];
+
+	[PrivateSentrySDKOnly setTrace:trace_id spanId:span_id];
+}
+
 void CocoaSDK::init() {
 	[PrivateSentrySDKOnly setSdkName:@"sentry.cocoa.godot"];
 
--- a/src/sentry/native/native_sdk.cpp
+++ b/src/sentry/native/native_sdk.cpp
@@ -382,6 +382,15 @@ void NativeSDK::remove_attribute(const String &p_name) {
 	sentry_remove_attribute(p_name.utf8());
 }
 
+void NativeSDK::set_trace(const String &p_trace_id, const String &p_parent_span_id) {
+	ERR_FAIL_COND(p_trace_id.is_empty());
+	if (p_parent_span_id.is_empty()) {
+		sentry_set_trace(p_trace_id.utf8(), NULL);
+	} else {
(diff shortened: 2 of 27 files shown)


src/sentry/cocoa/cocoa_sdk.mm
```diff
@@ -287,6 +287,17 @@ void _add_default_attachments(objc::SentryScope *p_scope) {
...
 }
 
+void CocoaSDK::set_trace(const String &p_trace_id, const String &p_parent_span_id) {
+	ERR_FAIL_COND(p_trace_id.is_empty());
+
+	objc::SentryId *trace_id = [[objc::SentryId alloc] initWithUUIDString:string_to_objc(p_trace_id)];
+	objc::SentrySpanId *span_id = p_parent_span_id.is_empty()
+			? [[objc::SentrySpanId alloc] init]
+			: [[objc::SentrySpanId alloc] initWithValue:string_to_objc(p_parent_span_id)];
+
```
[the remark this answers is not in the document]
limbonaut: From
https://github.com/getsentry/sentry-cocoa/blob/eddca8a95acd29d54ca2b84f3f3340213b0fed18/Sources/Sentry/Public/SentryId.h#L22-L25

>/// Creates a @c SentryId from a 32 character hexadecimal string without dashes such as
/// “12c2d058d58442709aa2eca08bf20986” or a 36 character hexadecimal string such as such as
/// “12c2d058-d584-4270-9aa2-eca08bf20986”.

limbonaut:
Sample issue: https://sentry-sdks.sentry.io/issues/7473696077/events/4697d843d8184af383a60b0ac3f0843a/

mujacica (approved):

src/sentry/native/native_sdk.cpp
```diff
@@ -382,6 +382,15 @@ void NativeSDK::remove_attribute(const String &p_name) {
 	sentry_remove_attribute(p_name.utf8());
 }
 
+void NativeSDK::set_trace(const String &p_trace_id, const String &p_parent_span_id) {
+	ERR_FAIL_COND(p_trace_id.is_empty());
+	if (p_parent_span_id.is_empty()) {
+		sentry_set_trace(p_trace_id.utf8(), NULL);
```
jpnurmi, line 388: Does it do something silly/wrong with a non-null but empty string?

jpnurmi: Btw, just asking in case there's anything to fix or improve in sentry-native :)

jpnurmi (approved):

src/sentry/native/native_sdk.cpp
limbonaut: Hmm, doesn't seem so. It's just a difference between serializing as `null` or as empty string. Now that you point it out, the branch doesn't seem to have any value - I'll remove it.

limbonaut: On the second look, according to docs:
https://develop.sentry.dev/sdk/foundations/transport/event-payloads/contexts/#trace-context
`parent_span_id` is optional and should be either null or a 16-char hex.

And relay seem to treat it as error:
https://github.com/getsentry/relay/blob/2061055ea0e7a82266ba4c90205c416daa04875d/relay-event-schema/src/protocol/contexts/trace.rs#L231

PR #713, opened 2026-05-18
feat(dotnet): Propagate default attachments into .NET layer

### LINKED ISSUE #651: Synchronize default attachments with .NET layer
**[GA blocker]**

On the native side, the log/screenshot/view-hierarchy files are added as file attachments on the native scope at `init()`, and `ScreenshotProcessor` / `ViewHierarchyProcessor` (registered via `process_event.cpp`) refresh those files per event so the attachments stay current. Events captured by the .NET SDK skip `sentry::process_event()` entirely, so neither the refresh nor the attachment wiring reaches them.

Work: hook a .NET-side event processor that triggers the native processors across interop (to refresh the files at capture time) and attaches the three files to the .NET event. Use the already-mirrored `AttachScreenshot` / `AttachSceneTree` / `AttachLog` flags.

limbonaut:
This PR propagates Godot SDK's default attachments (log, screenshot and view hierarchy) into the .NET layer so events captured from .NET include the same attachments as those captured from the native layer. These attachments are also automatically restored after the `ClearAttachments()` calls.
- Closes #651 
- Depends on #710

### DIFF
--- a/project/addons/sentry/dotnet/Sentry.Godot/Internal/GodotSdkIntegration.cs
+++ b/project/addons/sentry/dotnet/Sentry.Godot/Internal/GodotSdkIntegration.cs
@@ -48,6 +48,13 @@ internal sealed class GodotSdkIntegration : ISdkIntegration
             }
         }
 
+        AddDefaultAttachments(scope);
+    }
+
+    private static void AddDefaultAttachments(Scope scope)
+    {
+        // Don't propagate these scope changes to the native layer.
+        using var _ = new GodotScopeObserver.SyncGuard();
         foreach (SentryAttachment att in Sentry.Godot.SentrySdk.CurrentOptions!.DefaultAttachments)
         {
             scope.AddAttachment(att);
--- a/project/addons/sentry/dotnet/Sentry.Godot/Internal/GodotSdkIntegration.cs
+++ b/project/addons/sentry/dotnet/Sentry.Godot/Internal/GodotSdkIntegration.cs
@@ -47,17 +47,5 @@ internal sealed class GodotSdkIntegration : ISdkIntegration
                 scope.Contexts.App.Version = appVersion;
             }
         }
-
-        AddDefaultAttachments(scope);
-    }
-
-    private static void AddDefaultAttachments(Scope scope)
-    {
-        // Attachments came from native defaults, so skip syncing them back.
-        using var _ = new GodotScopeObserver.SyncGuard();
-        foreach (SentryAttachment att in Sentry.Godot.SentrySdk.CurrentOptions!.DefaultAttachments)
-        {
-            scope.AddAttachment(att);
-        }
     }
 }
--- a/project/addons/sentry/dotnet/Sentry.Godot/Internal/GodotSdkIntegration.cs
+++ b/project/addons/sentry/dotnet/Sentry.Godot/Internal/GodotSdkIntegration.cs
@@ -47,5 +47,10 @@ internal sealed class GodotSdkIntegration : ISdkIntegration
                 scope.Contexts.App.Version = appVersion;
             }
         }
+
+        foreach (SentryAttachment att in Sentry.Godot.SentrySdk.CurrentOptions!.DefaultAttachments)
+        {
+            scope.AddAttachment(att);
+        }
     }
 }
--- a/project/addons/sentry/dotnet/Sentry.Godot/Internal/GodotSdkIntegration.cs
+++ b/project/addons/sentry/dotnet/Sentry.Godot/Internal/GodotSdkIntegration.cs
@@ -53,7 +53,7 @@ internal sealed class GodotSdkIntegration : ISdkIntegration
 
     private static void AddDefaultAttachments(Scope scope)
     {
-        // Don't propagate these scope changes to the native layer.
+        // Attachments came from native defaults, so skip syncing them back.
         using var _ = new GodotScopeObserver.SyncGuard();
         foreach (SentryAttachment att in Sentry.Godot.SentrySdk.CurrentOptions!.DefaultAttachments)
         {
--- a/project/addons/sentry/dotnet/Sentry.Godot/Interop/GodotScopeObserver.cs
+++ b/project/addons/sentry/dotnet/Sentry.Godot/Interop/GodotScopeObserver.cs
@@ -79,19 +79,6 @@ internal class GodotScopeObserver : IScopeObserver
 
     public void ClearAttachments()
     {
-        if (IsSyncing || SentrySdk.InLocalScope)
-        {
-            return;
-        }
-        using var _ = new SyncGuard();
-
         // TODO: forward to native attachment API once wired through NativeBridge.
-
-        // Restore default attachments from the current options.
-        // These are maintained by the native layer and include screenshot, view hierarchy, and log file.
-        foreach (SentryAttachment att in Sentry.Godot.SentrySdk.CurrentOptions!.DefaultAttachments)
-        {
-            Sentry.Godot.SentrySdk.ConfigureScope(scope => scope.AddAttachment(att));
-        }
     }
 }
--- a/project/addons/sentry/dotnet/Sentry.Godot/Interop/GodotScopeObserver.cs
+++ b/project/addons/sentry/dotnet/Sentry.Godot/Interop/GodotScopeObserver.cs
@@ -79,6 +79,19 @@ internal class GodotScopeObserver : IScopeObserver
 
     public void ClearAttachments()
     {
+        if (IsSyncing || SentrySdk.InLocalScope)
+        {
+            return;
+        }
+        using var _ = new SyncGuard();
+
         // TODO: forward to native attachment API once wired through NativeBridge.
+
+        // Restore default attachments from the current options.
+        // These are maintained by the native layer and include screenshot, view hierarchy, and log file.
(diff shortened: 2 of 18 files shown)


src/sentry/dotnet/csharp_interop.cpp
```diff
@@ -129,6 +138,19 @@ struct NativeOptions {
...
+struct NativeArray {
+	void *ptr;
+	int32_t count;
+};
+
+CSHARP_EXPORT void csharp_interop_free_array(void *p_array) {
+	if (p_array) {
+		Memory::free_static(p_array, true);
+	}
+}
```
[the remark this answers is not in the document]
limbonaut: memdelete_arr uses the same alloc_static/free_static pair internally and accounts for padding:
https://github.com/godotengine/godot-cpp/blob/e83fd0904c13356ed1d4c3d09f8bb9132bdc6b77/src/core/memory.cpp#L91
This method is only used for trivially-destructible types.

src/sentry/dotnet/process_default_attachments.cpp
```diff
@@ -0,0 +1,24 @@
...
+	// TODO: Replace this with the full .NET event pipeline when it lands.
+	sentry::logging::print_debug("Processing default attachments...");
+	Ref<DisabledEvent> dummy_event;
+	dummy_event.instantiate();
+	for (const Ref<SentryEventProcessor> &processor : SENTRY_OPTIONS()->get_event_processors()) {
+		dummy_event = processor->process_event(dummy_event);
+		if (dummy_event.is_null()) {
+			return;
+		}
+	}
```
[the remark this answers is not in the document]
limbonaut: The architecture assumes that the processor will return either the same event or null. In practice, this should never happen.

limbonaut:
@sentry review

project/addons/sentry/dotnet/Sentry.Godot/Interop/GodotScopeObserver.cs
```diff
@@ -79,6 +79,19 @@ public void AddAttachment(SentryAttachment attachment)
 
     public void ClearAttachments()
     {
+        if (IsSyncing || SentrySdk.InLocalScope)
+        {
+            return;
+        }
+        using var _ = new SyncGuard();
+
         // TODO: forward to native attachment API once wired through NativeBridge.
```
bitsandfoxes:
```suggestion
```

project/addons/sentry/dotnet/Sentry.Godot/Internal/GodotSdkIntegration.cs
```diff
@@ -47,5 +47,17 @@ private static void ApplyScopeChanges(Scope scope)
...
         }
+
+        AddDefaultAttachments(scope);
+    }
+
+    private static void AddDefaultAttachments(Scope scope)
+    {
+        // Attachments came from native defaults, so skip syncing them back.
+        using var _ = new GodotScopeObserver.SyncGuard();
+        foreach (SentryAttachment att in Sentry.Godot.SentrySdk.CurrentOptions!.DefaultAttachments)
```
bitsandfoxes: I assume `DefaultAttachments` get constructed during `Init`? So users can opt-in/out of whatever it is that they want to have attached?

project/addons/sentry/dotnet/Sentry.Godot/Interop/GodotScopeObserver.cs
```diff
@@ -79,6 +79,19 @@ public void AddAttachment(SentryAttachment attachment)
...
+
+        // Restore default attachments from the current options.
+        // These are maintained by the native layer and include screenshot, view hierarchy, and log file.
+        foreach (SentryAttachment att in Sentry.Godot.SentrySdk.CurrentOptions!.DefaultAttachments)
+        {
+            Sentry.Godot.SentrySdk.ConfigureScope(scope => scope.AddAttachment(att));
+        }
```
bitsandfoxes: `ClearAttachments` is not clearing anything but adding the default attachments again?

bitsandfoxes (commented):
I'm not sure I agree with the concept of `DefaultAttachments`. Screenshots and view hierarchy are at-the-moment snapshots of the game's state and ideally live inside their respective integration. These should not need to get synched, as these are captured on an event basis and possibly out-of date in the next frame anyway.
And logs is "just" a path for a file to be picked up.

project/addons/sentry/dotnet/Sentry.Godot/Internal/GodotSdkIntegration.cs
limbonaut: Default attachments are internal concept, not user-visible. These are attachments that are added by native integrations, namely screenshot, view hierarchy and log file. And they are gated by respective options.

They are added here:
https://github.com/getsentry/sentry-godot/blob/05c9c7b97c8e6ee97a32eeb5349d52b3c6d19364/src/sentry/sentry_sdk.cpp#L167-L170
https://github.com/getsentry/sentry-godot/blob/05c9c7b97c8e6ee97a32eeb5349d52b3c6d19364/src/sentry/sentry_sdk.cpp#L355-L387

project/addons/sentry/dotnet/Sentry.Godot/Interop/GodotScopeObserver.cs
limbonaut: It's not implemented yet, so I'd rather keep the note.

project/addons/sentry/dotnet/Sentry.Godot/Interop/GodotScopeObserver.cs
limbonaut: The goal of this PR is not to synchronize scope attachments between layers, but rather connect .NET layer to native integrations which create screenshot and view hierarchy files (screenshot is created once per frame on demand). Historically, it was implemented using file-based attachments. When an event is created, the event processors on the native side update screenshot (once per frame) and build view hierarchy. The log file path doesn't change, so it's just added as attachment during init.
Internally, these attachments are called "default" but they are not user-visible through API.

limbonaut:
> Screenshots and view hierarchy are at-the-moment snapshots of the game's state and ideally live inside their respective integration.

Native integrations add two event processors that handle screenshot and view hierarchy. These integrations maintain two attachment files that are updated as needed. See:
- https://github.com/getsentry/sentry-godot/blob/05c9c7b97c8e6ee97a32eeb5349d52b3c6d19364/src/sentry/processing/screenshot_processor.cpp#L17-L94
- https://github.com/getsentry/sentry-godot/blob/05c9c7b97c8e6ee97a32eeb5349d52b3c6d19364/src/sentry/processing/view_hierarchy_processor.cpp#L13-L52

I'm open to suggestions on how we can improve our native pipeline for screenshots and view hierarchy. This PR simply bridges what we currently have with the .NET layer.

jpnurmi:
I think I might have asked this before, but is there anything we could do in upstream SDKs to make it easier to manage those special attachments? What if we marked them as internal/persistent/sticky or such, and stored them in a different attachment list hidden from the users? Would something like that help?

bitsandfoxes:
My bad, I missed the point of files based attachments and the role of  `DefaultAttachmentsProcessor.cs` and `process_default_attachments.cpp`. That makes a lot more sense now. Thanks!

I still don't think these should be added as attachments on the scope. I'm thinking of something like:
1. Set `options.AttachScreenshot = true;`
2. `Init` adds an event processor that adds screenshot attachments to events
3. The `ScreenshotEventProcessor` adds the pure file attachment to the event

Same for view hierarchy and logs. This way the SDK does not have to handle a list of what it considers "default" attachments, or make them sticky/persistent, or figure out how to reapply them when the attachments on the scope get cleared.

The `DefaultAttachmentProcessor` still needs to be called as the very first thing in the processing pipeline. If native pipeline looks similar, it took the same options, same processors, it creates the correct file attachments, and the C# part can pick them up.

limbonaut:
> What if we marked them as internal/persistent/sticky

@jpnurmi That would certainly help, but I don't think it's needed since, as Stefan pointed out, integration attachments could be added in a processor via hint, as long as upstream sdks support such mechanism. But I'm not sure if we can add attachments via before-send callback in sentry-native, though.

project/addons/sentry/dotnet/Sentry.Godot/Interop/NativeBridge.cs
```diff
@@ -440,6 +462,48 @@ public static void ApplyNativeOptionsDefaults(SentryGodotOptions opts)
...
+                },
+                content: new FileAttachmentContent(path),
+                fileName: String.IsNullOrEmpty(filename)
+                        ? System.IO.Path.GetFileName(path)
+                        : filename,
+                contentType: contentType
+            ));
+        }
+        csharp_interop_free_array(result.Ptr);
+    }
```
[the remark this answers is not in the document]
limbonaut: Added automatic disposal instead.

limbonaut:
Thanks, @bitsandfoxes! I reworked the .NET side to add attachments within the processor, instead of keeping them on the scope, which does make a lot of sense. I don't think it's necessary to split it up into separate processors at this point, but we can keep that option open in case we'll need per-integration logic in the .NET layer in the future.

bitsandfoxes (approved):
Nice! :shipit:

jpnurmi (approved):

Issue #647, opened 2026-04-21
C# support follow-ups


limbonaut:
Tracks post-merge follow-ups for the experimental C# support added in PR #629. Items tagged **[GA blocker]** must close before we drop the experimental label; the rest are related bugs, testing, and docs work.

- Refs https://github.com/getsentry/sentry-godot/pull/629
- Refs https://github.com/getsentry/sentry-godot/issues/91 (keep open until C# support reaches GA)

PR #719, opened 2026-05-20
ref(dotnet): Distribute .NET libs as precompiled DLLs

limbonaut:
Distribute the .NET layer as precompiled assemblies (DLL + PDB + XML) so user projects don't compile our source, and it fixes issues with IntelliSense that couldn't find doc comments previously.

- Refs #647 

Managed sources moved from `project/addons/sentry/dotnet/Sentry.Godot/` to `src/sentry/dotnet/managed/` with a flatter project hierarchy. A root-level solution paired with directory-level instructions means `dotnet build` from the repo root builds every managed project and stages outputs into `project/addons/sentry/dotnet/lib/` ready for use. A new `dotnet-libs` job in the build workflow bundles those staged libs into the packaged artifact.


jpnurmi (approved):
Congrats on defeating MSBuild! 🥇

bitsandfoxes (approved):
Welcome to the club of pre-compiled packages! 🎉

PR #753, opened 2026-06-08
ref(anr): Align App Hang options with other gaming SDKs

limbonaut:
Aligns the app hang options with the other gaming SDKs, continuing the cross-SDK effort to standardize how detection of a frozen main thread is configured. The master toggle is renamed to `enable_app_hang_tracking` (`EnableAppHangTracking` in the C# layer), and the timeout is now expressed in milliseconds as `app_hang_timeout_ms` to match the units used elsewhere. The former `app_hang_tracking` and `app_hang_timeout_sec` options remain as deprecated aliases that forward to the new ones, and a Project Settings migration rewrites existing persisted values (converting the timeout from seconds to milliseconds) so projects upgrade without manual changes. 
- Waiting for #749 
- Refs https://github.com/getsentry/team-gdx/issues/166

PR #759, opened 2026-06-12
ref(options): Group Godot logger options under `options.godot_logger`

### LINKED ISSUE #758: Group Godot logger options under `options.godot_logger`
The flat `logger_*` options on `SentryOptions` form a single coherent feature: the SDK's integration with Godot's logging system, which captures engine, script, and shader errors as well as printed messages, and routes them to Sentry as events, breadcrumbs, or logs. Keeping them flat makes the root options cluttered, and the `logger_*` prefix is easy to confuse with `SentrySDK.logger`, Sentry's structured-logging API, which is an unrelated feature.

Group these options into a dedicated object accessed as `options.godot_logger`, following how other Sentry SDKs structure feature-specific option groups (for example, the `Logs` and `Metrics` option classes in sentry-java). The old flat names should remain as deprecated aliases until the next major release, and persisted project settings should be migrated automatically.

This better land before the stable 2.0.0 release. The 2.0 release already restructures the options surface (the Android options group, the app hang renames), so shipping the logger grouping alongside them gives users a single coherent migration instead of breaking changes spread across releases. The C#/.NET layer also ships its first stable API with 2.0, and doing this now means it never exposes the flat names at all.

limbonaut:
This PR groups the flat `logger_*` options into a new `godot_logger` options object on `SentryOptions`. These options configure the SDK's integration with Godot's logging system, which captures engine errors, script errors, and printed messages as Sentry events, breadcrumbs, and logs. The new structure follows how other Sentry SDKs group feature-specific options (for example, the `Logs` and `Metrics` option groups in sentry-java), shortens the option names, and removes the ambiguity between this integration and `SentrySDK.logger`, Sentry's structured-logging API.

- Closes #758 
- Waiting for: 
  - #753
  - #749
- Docs https://github.com/getsentry/sentry-docs/pull/18297

## For reviewers

- The old flat `logger_*` properties remain as deprecated aliases that proxy to the new object and warn on use, following the same pattern as the earlier app hang renames. Removal is scheduled with the other deprecations (January 2027 or version 3.0).
- Two deliberate naming decisions: `include_source` became `include_source_context` (matches the Python SDK option), and `GodotLoggerEventMask` with its `MASK_*` constants intentionally stays on `SentryOptions` (namespace level in C#) so that future options outside the logger group can reference it.
- Project settings moved to `sentry/godot_logger/*`, with a schema v3 migration that renames persisted settings automatically; the demo project's settings are migrated as part of this PR.
- The C# layer mirrors the new structure as a clean break without `[Obsolete]` shims, since it has never been released as stable.

### DIFF
--- a/src/sentry/sentry_options.cpp
+++ b/src/sentry/sentry_options.cpp
@@ -291,7 +291,7 @@ Ref<SentryOptions> SentryOptions::create_from_project_settings() {
 }
 
 void SentryOptions::deprecated_set_logger_messages_as_breadcrumbs(bool p_enabled) {
-	WARN_DEPRECATED_MSG("The \"logger_messages_as_breadcrumbs\" option is deprecated. Set the MASK_MESSAGE flag in \"logger_breadcrumb_mask\" instead.");
+	WARN_DEPRECATED_MSG("The \"logger_messages_as_breadcrumbs\" option is deprecated. Set the MASK_MESSAGE flag in \"godot_logger.breadcrumb_mask\" instead.");
 	BitField<GodotLoggerEventMask> mask = godot_logger->get_breadcrumb_mask();
 	if (p_enabled) {
 		mask.set_flag(GodotLoggerEventMask::MASK_MESSAGE);
--- a/src/sentry/sentry_options.cpp
+++ b/src/sentry/sentry_options.cpp
@@ -430,7 +430,7 @@ void SentryOptions::_bind_methods() {
 	}
 
 	// DEPRECATED: These properties are deprecated and remain for compatibility reasons.
-	// TODO: Remove these after December 2026 or in version 3.0.
+	// TODO: Remove these in January 2027 or in version 3.0.
 	BIND_PROPERTY(SentryOptions, PropertyInfo(Variant::BOOL, "logger_messages_as_breadcrumbs"), deprecated_set_logger_messages_as_breadcrumbs, deprecated_is_logger_messages_as_breadcrumbs_enabled);
 	BIND_PROPERTY(SentryOptions, PropertyInfo(Variant::BOOL, "app_hang_tracking"), deprecated_set_app_hang_tracking, deprecated_get_app_hang_tracking);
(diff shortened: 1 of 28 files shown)


src/sentry/sentry_options.cpp
```diff
@@ -262,18 +290,15 @@ Ref<SentryOptions> SentryOptions::create_from_project_settings() {
...
 }
 
-bool SentryOptions::is_logger_messages_as_breadcrumbs_enabled() const {
-	// DEPRECATED: Scheduled for removal.
-	return logger_breadcrumb_mask.has_flag(GodotLoggerEventMask::MASK_MESSAGE);
-}
-
-void SentryOptions::set_logger_messages_as_breadcrumbs(bool p_enabled) {
+void SentryOptions::deprecated_set_logger_messages_as_breadcrumbs(bool p_enabled) {
 	WARN_DEPRECATED_MSG("The \"logger_messages_as_breadcrumbs\" option is deprecated. Set the MASK_MESSAGE flag in \"logger_breadcrumb_mask\" instead.");
```
JoshuaMoelans: `logger_breadcrumb_mask` is also deprecated
```suggestion
	WARN_DEPRECATED_MSG("The \"logger_messages_as_breadcrumbs\" option is deprecated. Set the MASK_MESSAGE flag in \"godot_logger.breadcrumb_mask\" instead.");
```

JoshuaMoelans (approved):
LGTM, just two small fixes
- in https://github.com/getsentry/sentry-godot/blob/ref/godot-logger-renames/src/sentry/godot_error_types.h#L31 it still references `logger_event_mask` ; should be `godot_logger.event_mask`

src/sentry/sentry_options.cpp
limbonaut: Nice catch, thanks!

limbonaut:
- Docs PR https://github.com/getsentry/sentry-docs/pull/18297

PR #790, opened 2026-06-29
ci(packaging): Bundle license into addon and demo packages

limbonaut:
The packaging script now bundles the project's `LICENSE.md` into the released archives so every distribution carries its license terms: the addon package gets it under `addons/sentry/`, and the demo-project package gets it both at the archive root and under `addons/sentry/` so the addon stays licensed when it is installed or copied out on its own from the demo project.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Other changes
+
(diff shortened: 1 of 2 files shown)


limbonaut:
For team: You can check in CI - see [Package job => List contents step](https://github.com/getsentry/sentry-godot/actions/runs/28378395210/job/84078216271?pr=790).

JoshuaMoelans (approved):

Issue #788, opened 2026-06-29
Migrate from `PrivateSentrySDKOnly` to `SentryObjCSDK.internal`


philprime:
[`sentry-cocoa` 9.19.0](<https://github.com/getsentry/sentry-cocoa/releases/tag/9.19.0>) introduces a new structured hybrid SDK API via `SentrySDK.internal` (Swift) and `SentryObjCSDK.internal` (ObjC). The old `PrivateSentrySDKOnly` class is deprecated and will be removed in the next major version.

**Implementation PR:** getsentry/sentry-cocoa#8097<br>**Design doc:** [SENTRY-INTERNAL-API.md](<https://github.com/getsentry/sentry-cocoa/blob/main/develop-docs/SENTRY-INTERNAL-API.md>)

## Benefits

* **Eliminates internal header imports.** Drop `PrivateSentrySDKOnly.h` and `Sentry-Swift.h`. The new API is fully public via `<SentryObjC/SentryObjC.h>`.
* **Proper debug image lookup API.** `SentryBinaryImageCache` (internal class accessed directly) is replaced by `[[SentryObjCSDK internal] debug]` which accepts `NSArray<NSNumber *>` addresses directly — no string conversion needed.
* **SPM / xcframework compatible.** Works with binary distribution — no exposed private headers.

## Import Change

Godot's iOS code is Objective-C++ (`.mm`) — migrate to the `SentryObjC` target:

```objc
// Before
#import <Sentry/PrivateSentrySDKOnly.h>
#import <Sentry/Sentry-Swift.h>  // in cocoa_includes.h

// After — single import, no internal headers
#import <SentryObjC/SentryObjC.h>
```

### SDK Name

| Before | After |
| -- | -- |
| `[PrivateSentrySDKOnly setSdkName:@"sentry.cocoa.godot"]` | `[SentryObjCSDK internal].sdk.name = @"sentry.cocoa.godot"` |

### Tracing

| Before | After |
| -- | -- |
| `[PrivateSentrySDKOnly setTrace:trace_id spanId:span_id]` | `[[SentryObjCSDK internal] setTrace:trace_id spanId:span_id]` |

**Note:** If `trace_id` and `span_id` are currently `SentryId` / `SentrySpanId` (Swift types from `Sentry-Swift.h`), replace with `SentryObjCId` / `SentryObjCSpanId`:

```objc
SentryObjCId *traceId = [[SentryObjCId alloc] initWithUUIDString:traceString];
[...]

PR #791, opened 2026-06-30
ref(cocoa): Replace workaround with internal API for debug images lookup

limbonaut:
Replace current workaround for debug image lookup in Cocoa backend with internal API.

- Refs #788


bitsandfoxes (approved):
ahh.. nice! so pretty!

PR #799, opened 2026-07-06
WIP: Add support for current scopes in GDScript

limbonaut:
#skip-changelog for now

PR #808, opened 2026-07-10
feat: Expand event object API

limbonaut:
Add `set_context()`, `set_user()`, `set_fingerprint()` to `SentryEvent` API.
This work was done originally in #799, but since I'm sacking that PR, we can keep the good bits.

### DIFF
--- a/src/sentry/disabled/disabled_event.h
+++ b/src/sentry/disabled/disabled_event.h
@@ -50,6 +50,7 @@ public:
 	virtual void remove_tag(const String &p_key) override {}
 	virtual String get_tag(const String &p_key) override { return ""; }
 
+	virtual void set_context(const String &p_key, const Dictionary &p_value) override {}
 	virtual void merge_context(const String &p_key, const Dictionary &p_value) override {}
 
 	virtual void add_exception(const Exception &p_exception) override {}
--- a/src/sentry/disabled/disabled_event.h
+++ b/src/sentry/disabled/disabled_event.h
@@ -16,8 +16,6 @@ private:
 	String release;
 	String dist;
 	String environment;
-	Ref<SentryUser> user;
-	PackedStringArray fingerprint;
 
 protected:
 	static void _bind_methods() {}
@@ -52,9 +50,9 @@ public:
 	virtual void remove_tag(const String &p_key) override {}
 	virtual String get_tag(const String &p_key) override { return ""; }
 
-	virtual void set_user(const Ref<SentryUser> &p_user) override { user = p_user; }
+	virtual void set_user(const Ref<SentryUser> &p_user) override {}
 
-	virtual void set_fingerprint(const PackedStringArray &p_fingerprint) override { fingerprint = p_fingerprint; }
+	virtual void set_fingerprint(const PackedStringArray &p_fingerprint) override {}
 
 	virtual void set_context(const String &p_key, const Dictionary &p_value) override {}
 	virtual void merge_context(const String &p_key, const Dictionary &p_value) override {}
--- a/src/sentry/disabled/disabled_event.h
+++ b/src/sentry/disabled/disabled_event.h
@@ -17,6 +17,7 @@ private:
 	String dist;
 	String environment;
 	Ref<SentryUser> user;
+	PackedStringArray fingerprint;
 
 protected:
 	static void _bind_methods() {}
@@ -52,6 +53,8 @@ public:
 	virtual String get_tag(const String &p_key) override { return ""; }
 
 	virtual void set_user(const Ref<SentryUser> &p_user) override { user = p_user; }
+
+	virtual void set_fingerprint(const PackedStringArray &p_fingerprint) override { fingerprint = p_fingerprint; }
 
 	virtual void set_context(const String &p_key, const Dictionary &p_value) override {}
 	virtual void merge_context(const String &p_key, const Dictionary &p_value) override {}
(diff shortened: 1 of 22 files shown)


src/sentry/native/native_event.cpp
```diff
@@ -175,6 +175,50 @@ String NativeEvent::get_tag(const String &p_key) {
...
+	}
+	if (!p_user->get_email().is_empty()) {
+		sentry_value_set_by_key(user_data, "email",
+				sentry_value_new_string(p_user->get_email().utf8()));
+	}
+	if (!p_user->get_ip_address().is_empty()) {
+		sentry_value_set_by_key(user_data, "ip_address",
+				sentry_value_new_string(p_user->get_ip_address().utf8()));
+	}
+	sentry_value_set_by_key(native_event, "user", user_data);
```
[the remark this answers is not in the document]
limbonaut: is_valid() checks if reference is not null - it's opposite of is_null() check.

src/sentry/native/native_event.cpp
```diff
@@ -175,6 +175,50 @@ String NativeEvent::get_tag(const String &p_key) {
...
+	if (p_user.is_null()) {
+		sentry_value_remove_by_key(native_event, "user");
+		return;
+	}
+
+	sentry_value_t user_data = sentry_value_new_object();
+	if (!p_user->get_id().is_empty()) {
+		sentry_value_set_by_key(user_data, "id",
+				sentry_value_new_string(p_user->get_id().utf8()));
+	}
```
jpnurmi, line 188: creating an empty user object loses the `user.id` pre-populated from the installation ID - should it be retained?
- https://github.com/getsentry/sentry-native/pull/1661

limbonaut: Godot SDK maintains its own installation id:
https://github.com/getsentry/sentry-godot/blob/cdd14bd517f8b451dc292c37e422d865e1570eba/src/sentry/sentry_user.cpp#L11-L22
If a dev sets ID as empty explicitly, I think we should honor it.

jpnurmi: Oh ok, then the installation ID in sentry-native is irrelevant 🙂

jpnurmi (approved):

src/sentry/native/native_event.cpp
```diff
@@ -175,6 +175,50 @@ String NativeEvent::get_tag(const String &p_key) {
...
+
+void NativeEvent::set_context(const String &p_key, const Dictionary &p_value) {
+	ERR_FAIL_COND_MSG(p_key.is_empty(), "Sentry: Can't set context with an empty key.");
+	sentry_value_t contexts = sentry_value_get_by_key(native_event, "contexts");
+	if (sentry_value_is_null(contexts)) {
+		contexts = sentry_value_new_object();
+		sentry_value_set_by_key(native_event, "contexts", contexts);
+	}
+	sentry_value_set_by_key(contexts, p_key.utf8(), sentry::native::variant_to_sentry_value(p_value));
+}
```
JoshuaMoelans, line 220: in native we recently added `update_context` (https://github.com/getsentry/sentry-native/pull/1835), do you think it makes sense to expose that one too? It provides this

```c
ctx={a:0, b:1, c:2}
sentry_update_context(ctx, {a:-1, d:3})
// ctx becomes {a:-1, b:1, c:2, d:3}
```
next to the `set_context` API, which is overwriting everything

src/sentry/disabled/disabled_event.h
```diff
@@ -50,6 +52,11 @@ class DisabledEvent : public SentryEvent {
 	virtual void remove_tag(const String &p_key) override {}
 	virtual String get_tag(const String &p_key) override { return ""; }
 
+	virtual void set_user(const Ref<SentryUser> &p_user) override { user = p_user; }
+
+	virtual void set_fingerprint(const PackedStringArray &p_fingerprint) override { fingerprint = p_fingerprint; }
+
+	virtual void set_context(const String &p_key, const Dictionary &p_value) override {}
```
JoshuaMoelans: q: is there a reason we have function bodies for `set_user` and `set_fingerprint` but not for `set_context`?

JoshuaMoelans (approved):
lgtm, just one open question (and the note left before about the other context API in native)

src/sentry/native/native_event.cpp
limbonaut: Actually, we already expose a similar method, but named differently:

https://github.com/getsentry/sentry-godot/blob/cdd14bd517f8b451dc292c37e422d865e1570eba/src/sentry/native/native_event.cpp#L222-L225
https://github.com/getsentry/sentry-godot/blob/69c365882f582a2e290ba21b87c757b8879814df/src/sentry/native/native_event.cpp#L20-L47

src/sentry/disabled/disabled_event.h
limbonaut: Generally, it's a noop event used when the SDK is not currently active or supported (so the user code doesn't break). But to answer your question, I probably felt lazy to do a consistent implementation 😅
I'll clear the bodies seeing that we don't have the matching getters exposed.

src/sentry/native/native_event.cpp
JoshuaMoelans: ah nice, at some point we might want to migrate these in-engine-sdk implementations of object/scope/context merging to just be an API surface in `sentry-native` (since for UE we were considering adding the same), but no need to expose `update_context` for now then 👌

limbonaut: And it looks like that event API is missing such: only `sentry_update_context` is exposed, no matching `sentry_event_update_context`.

PR #811, opened 2026-07-14
feat: Harden shutdown v2

### LINKED ISSUE #797: Using a Lambda for SentryOptions callbacks results in Sentry crashing on exiting
### How do you use Sentry?

Sentry SaaS

### Sentry SDK version

2.0.1

### How did you install the SDK?

AssetLib

### Godot version

v4.7.stable.official [5b4e0cb0f]

### Which platform?

Windows, Mac, Linux

### How to reproduce

[sentry-crash-on-exit.zip](https://github.com/user-attachments/files/29601612/sentry-crash-on-exit.zip)
Running the project and clicking on the quit button results in a crash. Using the non lambda function `_sentry_before_send` for the callback works. 
Probably the same holds true for all the `before_*` callbacks of SentryOptions.
I could reproduce the same issue on Linux, Mac and Windows but on Windows it's more silent since the crash only appears in the Event Viewer and not as a system notification.
Lambda Callables have is_custom() set to true while the normal method Callable has set it to false. Feels like custom Callables used as Sentry callbacks leads to the crash.

### Expected result

Either lambdas can be used as callbacks in SentryOptions or an error is thrown when assigning a custom Callable to a callback. 
If that is not possible, adding the caveat to the documentation of SentryOptions would help as well.

### Actual result

Godot crashes when quitting.

limbonaut:
The SDK can process events, including on background threads, while the engine is shutting down. During this time, it may access Godot machinery that is being torn down, such as the scene tree, display server, scripting runtime, and user callbacks such as `before_send` and friends. This PR makes shutdown safe by ending telemetry at a well-defined point: once the engine begins tearing down, the SDK closes every layer while the scripting runtime is still alive, before Godot tears down the systems SDK depends on. Closing also releases the user callbacks held in options, which fixes a crash on exit when `before_send` is a GDScript lambda (an anonymous inline function) due to an upstream issue.

- Fixes #797
- Supersedes #803

Since Godot doesn't provide a suitable signal to observe engine shutdown, we add a watcher to the scene tree as an internal node at the front. This ensures the node is freed as one of the last nodes in the scene tree. It's the closest point to scripting runtime destruction that we can observe.

#803 instead kept capture alive through shutdown, which required guarding every processing path, draining in-flight work, and adding an option to opt out. That window turned out to be worth very little: errors raised while Godot destroys the scene tree already go through the normal capture path, including `before_send`, and Godot 4.7 no longer routes late shutdown errors to the logging pipeline. Cutting telemetry at the scene tree boundary gives up almost no coverage and drops that complexity, along with the `enable_capture_during_shutdown` option.


src/sentry/engine_lifecycle/engine_lifecycle.cpp
```diff
@@ -0,0 +1,75 @@
...
+	SentrySceneTreeWatcher *watcher = memnew(SentrySceneTreeWatcher);
+	watcher->set_shutdown_callback(callable_mp_static(&_scene_tree_shutting_down));
+	// Add at the front so it is torn down after other scene tree nodes.
+	root->add_child(watcher, false, Node::INTERNAL_MODE_FRONT);
+}
+
+void _engine_ready() {
+	_singletons_ready.store(true, std::memory_order_release);
+	_add_scene_tree_watcher();
+}
```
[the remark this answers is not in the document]
limbonaut: It's a startup flag, not shutdown. Flipping it upon shutdown would redefine its meaning.

src/sentry/sentry_sdk.cpp
```diff
@@ -499,10 +513,10 @@ void SentrySDK::prepare_and_auto_initialize() {
...
 	switch (p_what) {
 		case NOTIFICATION_PREDELETE: {
-			if (godot_logger.is_valid()) {
-				OS::get_singleton()->remove_logger(godot_logger);
-				godot_logger.unref();
-			}
+			sentry::engine_lifecycle::remove_shutdown_callback(
+					callable_mp(this, &SentrySDK::_on_engine_shutdown));
+			// Fallback in case _on_engine_shutdown() did not run.
+			close();
```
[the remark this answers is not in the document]
limbonaut: It already does, because singleton teardown happens past script runtime lifetime. It's a fallback.

src/sentry/engine_lifecycle/engine_lifecycle.cpp
```diff
@@ -0,0 +1,74 @@
...
+namespace sentry::engine_lifecycle {
+
+void start_lifecycle_watch() {
+	if (_watch_started) {
+		return;
+	}
+	// Deferred: the engine isn't fully up yet.
+	callable_mp_static(&_add_scene_tree_watcher).call_deferred();
+	_watch_started = true;
+}
```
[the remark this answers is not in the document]
limbonaut: The flag can only cause a missed retry if something calls start_lifecycle_watch() a second time, and nothing does.

src/sentry/engine_lifecycle/engine_lifecycle.cpp
```diff
@@ -0,0 +1,74 @@
...
+void _add_scene_tree_watcher() {
+	SceneTree *tree = Object::cast_to<SceneTree>(Engine::get_singleton()->get_main_loop());
+	ERR_FAIL_NULL_MSG(tree, "Sentry: Failed to initialize engine lifecycle tracking - SceneTree is unavailable.");
+	Window *root = tree->get_root();
+	ERR_FAIL_NULL_MSG(root, "Sentry: Failed to initialize engine lifecycle tracking - root window is unavailable.");
+
+	SentrySceneTreeWatcher *watcher = memnew(SentrySceneTreeWatcher);
+	watcher->set_shutdown_callback(callable_mp_static(&_scene_tree_shutting_down));
+	// Add at the front so it is torn down after other scene tree nodes.
+	root->add_child(watcher, false, Node::INTERNAL_MODE_FRONT);
```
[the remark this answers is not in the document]
limbonaut: We don't support live extension reload, and Godot doesn't implement explicit UI controls over extensions like that, so in practice it gets to load only once. If that ever changes, we would have to adjust this.

src/sentry/sentry_options.cpp
```diff
@@ -37,6 +37,34 @@ void _define_setting(const godot::PropertyInfo &p_info, const godot::Variant &p_
...
+	constexpr uint32_t MAX_DEPTH = 4;
+	if (p_object == nullptr || p_depth > MAX_DEPTH) {
+		return;
+	}
+
+	TypedArray<Dictionary> properties = p_object->get_property_list();
+
+	// Clear callables on this object before recursing. Some nested properties can proxy back to their owner.
+	// Clearing the owner's callables first makes those proxy properties read back as invalid/unset,
+	// so we avoid setting them and emitting deprecation warnings.
```
jpnurmi: I found only 4 callables. The alternative is something like:
```cpp
void SentryOptions::release_callables() {
	before_send_log = Callable();
	before_send_metric = Callable();
	before_send = Callable();
	before_capture_screenshot = Callable();
}
```

Feel free to discard this as it's a matter of taste, but to me it feels so much simpler and more readable that I'd choose this at the cost of having to maintain it by hand. 🤓

limbonaut: The problem is that the price of forgetting to add it in here also invites crashing. And we can't catch it in tests,  because if a new callback is introduced, we'd have to add a regression test alongside. I'd rather avoid such footgun.

jpnurmi: Is it not possible to run the recursive check in tests instead of production? :upside_down_face: Nah, I don't mean to start arguing. Just a silly outsider observation from a C programmer used to such manual labor with very little convenience available. :sweat_smile:

jpnurmi (approved):
I like this simpler variant 👍

src/sentry/sentry_options.cpp
limbonaut: It might be possible to employ the same trick in testing. Callable bindings don't carry type information, making such tasks actually easier.

limbonaut: Good point! I moved it into tests and confirmed that it trips if a callback is not cleared.

jpnurmi: Sorry... one day the test will fail, and I'll live rent-free in your head. :smirk:

bitsandfoxes (approved):
Awesome iteration! Thanks! 🔥 :shipit:

JoshuaMoelans (approved):
nice work, LGTM

PR #826, opened 2026-07-24
fix(android): Missing GDScript frame lineno on Android in Godot 4.6+

limbonaut:
Fixes missing line numbers in GDScript frames on Android with Godot 4.6 and later. Flagged by new testing added in #825.

<details>
<summary>Test failure</summary>

</details>

Root cause: Godot 4.6 changed how `Variant::INT` is boxed when it crosses into Java. See commit [`b0cb297cde`](https://github.com/godotengine/godot/commit/b0cb297cde). We cannot rely on direct casts anymore.

This PR also hardens another place for potential failure, fixes a bug in the bridge-layer conversion methods and adds corresponding unit tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- Fixed missing line numbers in GDScript frames on Android with Godot 4.6 and later ([#826](https://github.com/getsentry/sentry-godot/pull/826))
+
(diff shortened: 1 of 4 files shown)


limbonaut:
I ran the tests locally with different Godot versions, and they pass now. The updated testing in #825 should catch such issues in the future.

jpnurmi (approved):

Issue #188, opened 2025-06-03
Add support for local scopes


limbonaut:
Allow capturing events with local scopes as in other Sentry SDKs.

Documentation:
- [Scopes in dev docs](https://develop.sentry.dev/sdk/telemetry/scopes/).
- [Scopes in Javascript](https://docs.sentry.io/platforms/javascript/enriching-events/scopes/).

Depends on:
- https://github.com/getsentry/sentry-native/issues/1226
- https://github.com/getsentry/sentry-native/pull/1248

limbonaut:
Docs PR: https://github.com/getsentry/sentry-docs/pull/18930

limbonaut:
Merged support for all platforms except Apple. Apple support depends on https://github.com/getsentry/sentry-cocoa/issues/8540.
Scoped attachments are added separately in #856, which crosses off two other tickets as well.

limbonaut:
Support for Apple platforms is tracked separately in #860.

PR #836, opened 2026-07-28
feat(scopes): Support scopes on WebAssembly

limbonaut:
Adds WebAssembly/JS support for scopes and `SentrySDK.with_scope()` in GDScript. See the https://github.com/getsentry/sentry-godot/pull/834 for additional context and implementation details.

- Stacked PR; root branch: https://github.com/getsentry/sentry-godot/pull/834
- Refs #188
- Docs https://github.com/getsentry/sentry-docs/pull/18930

### DIFF
--- a/src/sentry/javascript/javascript_scope.cpp
+++ b/src/sentry/javascript/javascript_scope.cpp
@@ -1,0 +1,100 @@
+#include "javascript_scope.h"
+
+#include "sentry/javascript/javascript_breadcrumb.h"
+#include "sentry/sentry_sdk.h"
+
+#include <godot_cpp/classes/json.hpp>
+#include <godot_cpp/core/error_macros.hpp>
+
+namespace sentry::javascript {
+
+void JavaScriptScope::set_context(const String &p_key, const Dictionary &p_value) {
+	ERR_FAIL_COND_MSG(p_key.is_empty(), "Sentry: Can't set context with an empty key.");
+	ERR_FAIL_COND(!js_obj);
+	js_bridge()->call("scopeSetContext", js_obj, p_key.utf8(), JSON::stringify(p_value).utf8());
+}
+
+void JavaScriptScope::set_tag(const String &p_key, const String &p_value) {
+	ERR_FAIL_COND_MSG(p_key.is_empty(), "Sentry: Can't set tag with an empty key.");
+	ERR_FAIL_COND(!js_obj);
+	js_obj->call("setTag", p_key.utf8(), p_value.utf8());
+}
+
+void JavaScriptScope::set_user(const Ref<SentryUser> &p_user) {
+	ERR_FAIL_COND(!js_obj);
+
+	if (p_user.is_null()) {
+		js_obj->call("setUser", nullptr);
+		return;
+	}
+
+	js_bridge()->call("scopeSetUser",
+			js_obj,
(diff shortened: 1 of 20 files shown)


src/sentry/javascript/javascript_sdk.cpp
```diff
@@ -111,6 +112,18 @@ static void before_send_metric_wasm_callback(int32_t *p_ids, int32_t p_len) {
 
 } // extern "C"
 
+namespace {
+
+// Returns the JS Scope object backing p_scope, or null if it's not backed by one (ie DisabledScope).
+// Scope creation can fall back to DisabledScope, and null falls back to capture without passed scope.
+inline JSObjectPtr _get_scope_object(const Ref<SentryScope> &p_scope) {
+	ERR_FAIL_COND_V(p_scope.is_null(), nullptr);
+	JavaScriptScope *js_scope = dynamic_cast<JavaScriptScope *>(p_scope->get_implementation());
```
jpnurmi: is there any godot alternative to `dynamic_cast`? could some game want `-fno-rtti`? or does godot/wasm perhaps already require RTTI?

limbonaut: Considering that we are compiling against Godot 4.5 baseline, Godot CPP already [pulls RTTI](https://github.com/godotengine/godot-cpp/blob/7e18e40d7591429f915035a7de7cf79457d555cc/include/godot_cpp/core/object.hpp#L133).

We could implement a light-weight system for pure C++ objects, but right now it would be mostly for performance. I can give it a shot for science :microscope: :mag:

limbonaut: So, I implemented type tagging in [497f94e](https://github.com/getsentry/sentry-godot/pull/836/commits/497f94e02023358a81391eea529193456e0b12cd) (see  `Castable`), and measured about 5x speeduip over dynamic_cast on my system.

src/sentry/javascript/javascript_scope.cpp
```diff
@@ -0,0 +1,85 @@
...
+			js_obj->call("setAttribute", p_name.utf8(), p_value.operator int64_t());
+		} break;
+		case Variant::Type::FLOAT: {
+			js_obj->call("setAttribute", p_name.utf8(), p_value.operator double());
+		} break;
+		case Variant::Type::STRING: {
+			js_obj->call("setAttribute", p_name.utf8(), p_value.operator String().utf8());
+		} break;
+		default: {
+			js_obj->call("setAttribute", p_name.utf8(), p_value.stringify().utf8());
```
jpnurmi, line 59: does this fallback work for [arrays](https://develop.sentry.dev/sdk/foundations/state-management/scopes/attributes/#complex-types)? the earlier PRs might have the same limitation so could be a follow-up if arrays are possible to support?

jpnurmi (approved):

src/sentry/javascript/javascript_scope.cpp
limbonaut: The API we expose doesn't support array-types for attributes anywhere in the chain yet (except .NET), and the unsupported types are stringified, so I think this code aligns well with the rest. This needs to be added in a separate PR, a ticket to track it: 
- https://github.com/getsentry/sentry-godot/issues/853

PR #845, opened 2026-07-30
fix(android): capture_event() returns an ID for discarded events

limbonaut:
On Android, `SentrySDK.capture_event()` returned the ID stored on the event object instead of the ID that sentry-java reported back from the capture call. The two differ whenever an event is dropped: when a `before_send` callback discards an event, sentry-java returns a nil ID of all zeros, but the plugin threw that value away and handed back the event's own ID. Callers were therefore given a valid-looking ID for an event that was never sent, with no way to tell the difference.
- Extracted from #834


bitsandfoxes (approved):
Whoa.. nice find!

JoshuaMoelans (approved):

PR #869, opened 2026-08-14
feat(options): Deprecate enable_logs and enable_metrics

### LINKED ISSUE #864: Make `enable_logs` option noop and drop it in the next major
Make `SentryOptions.enable_logs` a deprected noop in the next release and drop it in the next major alongside `enabled_metrics`.

limbonaut:
Deprecates the `enable_logs` and `enable_metrics` options on `SentryOptions`, together with the matching entries in Godot's Project Settings (the per-project configuration UI). Both options keep working exactly as they do today, so this only adds deprecation markers and respective warnings if those options are disabled.

- Resolves #864

These options are redundant because logs and metrics are sent only when a project uses `SentrySDK.logger`, `SentrySDK.metrics`, or opts into automatic log capture through `godot_logger.log_mask`.

The warning is attached to the scripting property binding rather than to the C++ setters, because the .NET layer writes its options back into the native options on every initialization and would otherwise warn on every C# project regardless of whether it touched the option. The Project Settings entries warn separately, and only when switched off, since the enabled default says nothing about user intent. The C# `EnableLogs` and `EnableMetrics` properties come from the .NET SDK rather than this repository, so they cannot be marked obsolete here and are unaffected.

### DIFF
--- a/src/sentry/sentry_options.cpp
+++ b/src/sentry/sentry_options.cpp
@@ -231,6 +231,14 @@ void SentryOptions::_load_project_settings(const Ref<SentryOptions> &p_options) 
 	p_options->enable_logs = ProjectSettings::get_singleton()->get_setting("sentry/options/enable_logs", p_options->enable_logs);
 	p_options->enable_metrics = ProjectSettings::get_singleton()->get_setting("sentry/options/enable_metrics", p_options->enable_metrics);
 
+	// Only a disabled setting is worth warning about: the enabled default carries no user intent.
+	if (!p_options->enable_logs) {
+		WARN_PRINT_ONCE("Sentry: The \"Enable Logs\" project setting is deprecated and will be removed in version 3.0. Logs are only sent if you use SentrySDK.logger or set \"sentry/godot_logger/logs\".");
+	}
+	if (!p_options->enable_metrics) {
+		WARN_PRINT_ONCE("Sentry: The \"Enable Metrics\" project setting is deprecated and will be removed in version 3.0. Metrics are only sent if you use SentrySDK.metrics.");
+	}
+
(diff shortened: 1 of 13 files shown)


src/sentry/sentry_options.cpp
tustanivsky: Since this is now a dead-end migration path should we update message here as well?

tustanivsky (approved):

src/sentry/sentry_options.cpp
limbonaut: Missed that one, thanks!

src/sentry/sentry_options.cpp
```diff
@@ -231,6 +231,14 @@ void SentryOptions::_load_project_settings(const Ref<SentryOptions> &p_options)
 	p_options->enable_logs = ProjectSettings::get_singleton()->get_setting("sentry/options/enable_logs", p_options->enable_logs);
 	p_options->enable_metrics = ProjectSettings::get_singleton()->get_setting("sentry/options/enable_metrics", p_options->enable_metrics);
 
+	// Only a disabled setting is worth warning about: the enabled default carries no user intent.
+	if (!p_options->enable_logs) {
+		WARN_PRINT_ONCE("Sentry: The \"Enable Logs\" project setting is deprecated and will be removed in version 3.0. Logs are only sent if you use SentrySDK.logger or set \"sentry/godot_logger/logs\".");
+	}
+	if (!p_options->enable_metrics) {
+		WARN_PRINT_ONCE("Sentry: The \"Enable Metrics\" project setting is deprecated and will be removed in version 3.0. Metrics are only sent if you use SentrySDK.metrics.");
```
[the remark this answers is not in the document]
limbonaut: you can't differentiate true from default value.

Issue #861, opened 2026-08-10
Add Spans API for GDScript


limbonaut:
Support adding and instrumenting spans in GDScript. API needs to include `start_span`, `with_span` etc.

* Span API docs: [https://develop.sentry.dev/sdk/telemetry/spans/](<https://develop.sentry.dev/sdk/telemetry/spans/>)
* Depends on getsentry/sentry-godot#188
* Also getsentry/sentry-godot#860

PR #892, opened 2026-08-26
feat(spans): Add SentrySDK.with_span()

limbonaut:
Adds `SentrySDK.with_span()`, a convenience wrapper over `start_span()` that starts an active span, calls a callable with that span, ends it when the callable returns, and hands back whatever the callable returned. It fits operations that live in a single synchronous callable, while `start_span()` stays the right choice when the span needs initial attributes, an explicit parent, an inactive mode, or a lifetime spanning several functions. Nested calls nest their spans, and the enclosing span and scope are restored on return. Split out of #863 so it can be reviewed on its own.

- Refs #861
- Docs PR https://github.com/getsentry/sentry-docs/pull/19210

Because the span ends the moment the callable returns, it only covers the callable's synchronous part. A callable that awaits hands control back at the first `await` and returns a coroutine state object rather than its eventual result, so the span ends early. The SDK detects that case and prints a one-time warning.


mujacica (approved):

jpnurmi (approved):

limbonaut:
- Docs PR: https://github.com/getsentry/sentry-docs/pull/19210

Issue #907, opened 2026-09-01
Trace propagation


limbonaut:
In the GDScript API, add support for trace propagation as described in specs:
- [Trace propagation](https://develop.sentry.dev/sdk/foundations/trace-propagation/) — carriers, `tracePropagationTargets`, strict continuation, `propagateTraceparent`, deferred sampling.
- [Dynamic sampling context](https://develop.sentry.dev/sdk/foundations/trace-propagation/dynamic-sampling-context/) — what belongs in `baggage` and when it freezes.
- [Span trace propagation](https://develop.sentry.dev/sdk/telemetry/spans/span-trace-propagation/) — API shape.
- [Scopes](https://develop.sentry.dev/sdk/foundations/state-management/scopes/) — the propagation context lives on a scope.

Expected API additions:
```gdscript
SentrySDK.start_new_trace(...)
SentrySDK.continue_trace(...)
SentrySDK.get_trace_headers(...)
SentryOptions.trace_propagation_targets
SentryOptions.propagate_traceparent
SentryOptions.strict_trace_continuation
```

limbonaut:
Complete in #915 (except Apple - support coming soon).

PR #909, opened 2026-09-01
feat(tracing): Add SentrySDK.start_new_trace()

limbonaut:
Adds `SentrySDK.start_new_trace()`[^1], which starts a trace unconnected to the one in progress so that telemetry captured afterwards is grouped separately from what came before. A game is a long-running process, and until now everything captured in a session shared the single trace the SDK starts at initialization. The new trace ID is minted in the SDK's own layer and routed through the existing internal `set_trace()` path, which keeps the cached trace context consistent with what the backend holds and needs no new backend interface.

- Refs #907

[^1]: [startNewTrace in trace propagation spec](https://develop.sentry.dev/sdk/foundations/trace-propagation/#startnewtrace)

### DIFF
--- a/doc_classes/SentrySDK.xml
+++ b/doc_classes/SentrySDK.xml
@@ -204,7 +204,7 @@
 			<description>
 				Starts a new trace, unconnected to the current one. Telemetry captured from this point on carries the new trace ID and no link back to the trace it replaces.
 				Use this to draw a clean trace boundary where one piece of work ends and unrelated work begins. Without it, everything captured in a session shares the trace the SDK starts at initialization.
-				[b]Note:[/b] A span started with [method SentrySDK.start_span] keeps the trace it began on, and so does the telemetry captured while that span is active. Start the new trace once the operation it measures has finished.
+				[b]Note:[/b] A span started with [method SentrySDK.start_span] keeps the trace it began on, and so does the telemetry captured while that span is active. The new trace takes effect once that span ends.
 				To learn more, visit [url=https://docs.sentry.io/concepts/key-terms/tracing/]Tracing documentation[/url].
 			</description>
 		</method>
--- a/doc_classes/SentrySDK.xml
+++ b/doc_classes/SentrySDK.xml
@@ -202,7 +202,7 @@
 		<method name="start_new_trace">
 			<return type="void" />
 			<description>
-				Starts a new trace, unconnected to the current one. Telemetry captured from this point on carries the new trace ID and no link back to the trace it replaces.
+				Starts a new trace with a new trace ID. A trace groups connected events and operations under one trace ID. A new trace begins a separate group in Sentry.
 				Use this to draw a clean trace boundary where one piece of work ends and unrelated work begins. Without it, everything captured in a session shares the trace the SDK starts at initialization.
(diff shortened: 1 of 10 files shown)


jpnurmi (approved):

doc_classes/SentrySDK.xml
```diff
@@ -199,6 +199,15 @@
 				Assigns user data. See [SentryUser].
 			</description>
 		</method>
+		<method name="start_new_trace">
+			<return type="void" />
+			<description>
+				Starts a new trace, unconnected to the current one. Telemetry captured from this point on carries the new trace ID and no link back to the trace it replaces.
```
bitsandfoxes: This reads confusing to me. What is "the current one". What happens to it when starting a new one? Does it get finshed? If the new one finishes, does the "current" one get restored?

doc_classes/SentrySDK.xml
```diff
@@ -199,6 +199,15 @@
...
+				Use this to draw a clean trace boundary where one piece of work ends and unrelated work begins. Without it, everything captured in a session shares the trace the SDK starts at initialization.
+				[b]Note:[/b] A span started with [method SentrySDK.start_span] keeps the trace it began on, and so does the telemetry captured while that span is active. The new trace takes effect once that span ends.
```
bitsandfoxes, line 207: Great usage explainer and sample. And thanks for noting down the potential footgun.

doc_classes/SentrySDK.xml
```diff
@@ -199,6 +199,15 @@
 				Assigns user data. See [SentryUser].
 			</description>
 		</method>
+		<method name="start_new_trace">
```
bitsandfoxes, line 202: i assume this was named after the trace utils in sentry-java/javascript? worth noting, that godot is the first of the gaming sdks to add this api.

src/sentry/sentry_sdk.cpp
```diff
@@ -425,6 +425,10 @@ void SentrySDK::remove_user() {
 	}
 }
 
+void SentrySDK::start_new_trace() {
+	set_trace(sentry::uuid::make_uuid_no_dashes(), String());
```
bitsandfoxes, line 429: is set_trace calling the scope observer? is that how that works for C#?

limbonaut: See #910

doc_classes/SentrySDK.xml
limbonaut: This name is straight from the dev docs. See https://develop.sentry.dev/sdk/foundations/trace-propagation/#startnewtrace
And yes, from the gaming SDKs, Godot is the first to add it.

doc_classes/SentrySDK.xml
limbonaut: Looks like it needs a better take. There is no finish or coming back - to end trace, you call it again and start a new one. The next paragraph elaborates on what it does and why.

mujacica (approved):

doc_classes/SentrySDK.xml
limbonaut: Updated in [d7415df](https://github.com/getsentry/sentry-godot/pull/909/commits/d7415df40f97aa7b207517e5084d243c5ed2117d)

PR #934, opened 2026-09-10
test(ios): Close Sentry before exiting integration tests

limbonaut:
Close Sentry after an iOS CLI command completes, before printing the exit marker and forcibly terminating the process. Forced termination bypasses normal SDK shutdown and can interrupt pending uploads. Centralizing cleanup also covers post-crash relaunches while keeping Sentry active during intentional crashes.

### DIFF
--- a/project/main.gd
+++ b/project/main.gd
@@ -7,6 +7,10 @@ func _ready() -> void:
 	SentrySDK.logger.info("Starting UI on %s", [OS.get_name()])
 
 	if await cli_commands.check_and_execute_cli():
+		if OS.get_name() == "iOS":
+			# Forced iOS termination skips normal shutdown,
+			# so flush pending events first.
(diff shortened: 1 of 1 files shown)


jpnurmi (commented):
Somehow the problem doesn't sound iOS-specific. 🤔

Or does it have something to do with how app-runner terminates apps on iOS, specifically? Do other platforms have graceful exit, allowing them to flush?

limbonaut:
@jpnurmi Godot has not "quit application" on iOS. The normal quit is no-op due to some platform specifics, so we employ process kill to end our run. Because we do that, it omits the normal flush happening on application termination. See https://docs.godotengine.org/en/stable/classes/class_scenetree.html#class-scenetree-method-quit

jpnurmi (approved):

